PR #1, opened 2020-12-04
Feat: character creation

rhydev:
- Added static class AccountStorage for dynamic character data
- Removed references to StaticAccountStorage
- Implemented character creation
- Added missing character and packet information for character appearance

### DIFF
--- a/MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
+++ b/MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
@@ -17,11 +17,7 @@ namespace MapleServer2.PacketHandlers.Login {
     public class CharacterManagementHandler : LoginPacketHandler {
         public override ushort OpCode => RecvOp.CHARACTER_MANAGEMENT;
 
-        private readonly IAccountStorage _accountStorage;
-        public CharacterManagementHandler(IAccountStorage accountStorage,ILogger<CharacterManagementHandler> logger) : base(logger) {
-
-            _accountStorage = accountStorage;
-        }
+        public CharacterManagementHandler(ILogger<CharacterManagementHandler> logger) : base(logger) {}
 
         public override void Handle(LoginSession session, PacketReader packet) {
             byte mode = packet.ReadByte();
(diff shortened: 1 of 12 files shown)


MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
```diff
@@ -89,46 +86,101 @@ public void HandleCreate(LoginSession session, PacketReader packet)
...
-                        Equips.Add(ItemSlot.HR, new Item(Convert.ToInt32(id)));
+                        // Hair Length/Position
+                        float backLength = BitConverter.ToSingle(packet.Read(4), 0);
+                        byte[] backPositionArray = packet.Read(24);
+                        float frontLength = BitConverter.ToSingle(packet.Read(4), 0);
+                        byte[] frontPositionArray = packet.Read(24);
+
+                        Equips.Add(ItemSlot.HR, new Item(Convert.ToInt32(id)) {
+                            Uid = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0),
+                            CreationTime = 1565575851,
```
eotetb, line 97: We probably don't want to hardcode the creation time. I recommend getting the current DateTime outside of the for loop and converting it to epoch time (that's what the hardcoded value is). There is a built in method that can allow you to do this conversion within the .NET framework.

https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimemilliseconds?view=netframework-4.8&viewFallbackFrom=netframework-4.0

You can easily double check and see if the value you're returning makes sense, there are multiple converters online, such as this https://www.epochconverter.com/

MapleServer2/Types/Player.cs
```diff
@@ -162,20 +162,36 @@ public static Player MaleDefault(long accountId, long characterId, string name =
...
+            stats.CurrentHp = new PlayerStat(0, 1000, 0);
+            stats.Spirit = new PlayerStat(100, 100, 100);
+            stats.Stamina = new PlayerStat(120, 120, 120);
+            stats.AtkSpd = new PlayerStat(120, 100, 130);
+            stats.MoveSpd = new PlayerStat(110, 100, 150);
+            stats.JumpHeight = new PlayerStat(110, 100, 130);
+
             var player = new Player
             {
+                AccountId = 0x1111111111111111,
```
eotetb, line 177: I don't know much about MS2's account ids, but there is a good chance that duplicate AccountIds are going to cause issues. We probably want to generate a new one for each account (I'm aware that the db isn't done at this point, if this is a placeholder, it might be worth leaving a TODO comment here). 

The same applies for other placeholder values we might be using right now, we don't want to forget about them.

eotetb (commented):
I see this has already been merged, but a few things in this PR caught my eye and may be worth taking a look at.

PR #3, opened 2020-12-08
Feat: skill tree

rhydev:
- Added handlers for REQUEST_SKILL_BOOK_TREE and JOB packets
- Added packet classes for SKILL_BOOK_TREE and JOB
- Added Skill type for storing skill data
- Revamped SkillTab type to store Skills
- Updated FIELD_ADD_USER packet's Skills info (no longer hardcoded)
- Added XML files for job and general skills
- Added JobSkillOrder static class to store exact skill id send order
- Various other small changes


MapleServer2/Constants/RecvOp.cs
```diff
@@ -1,152 +1,152 @@
...
+            USER_CHAT = 0x0011,
+            USER_SYNC = 0x0012,
+            EMOTION = 0x0013,
+            REQUEST_SERVER_LIST = 0x0014,
+            REQUEST_ITEM_CREATE = 0x0015,
+            ITEM_EQUIP = 0x0016,
+            REQUEST_ITEM_INVENTORY = 0x0017,
+            REQUEST_ITEM_STORAGE = 0x0018,
+            REQUEST_ITEM_USE = 0x0019,
+            ITEM_ExTRA_DATA = 0x001A,
```
AlanMorel, line 25: Uppercase x?

MapleServer2/PacketHandlers/Game/JobHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+                case 8: // Close Skill Tree
+                    HandleCloseSkillTree(session, packet);
+                    break;
+                case 9: // Save Skill Tree
+                    HandleSaveSkillTree(session, packet);
+                    break;
+            }
+        }
+
+        public void HandleCloseSkillTree(GameSession session, PacketReader packet) {
```
AlanMorel, line 26: This and the below method could be private

MapleServer2/PacketHandlers/Game/JobHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+            // Get skill tab to update
+            SkillTab skillTab = session.Player.SkillTabs[0]; // Get first skill tab only for now, uncertain of how to have multiple skill tabs
+
+            // Read skills
+            int count = packet.ReadInt(); // Number of skills
+            for (int i = 0; i < count; i++) {
+                // Read skill info
+                int id = packet.ReadInt(); // Skill id
+                short level = packet.ReadShort(); // Skill level
+                byte learned = packet.ReadByte(); // 00 if unlearned 01 if learned
```
AlanMorel, line 40: `learned` should be a boolean then if it's only ever going to have those two states

MapleServer2/PacketHandlers/Game/SkillBookTreeHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+
+        public void HandleOpen(GameSession session, PacketReader packet) {
+            session.Send(SkillBookTreePacket.Open(session.Player));
+        }
+
+        public void HandleSave(GameSession session, PacketReader packet) {
+            // Reads first 20 bytes, some kind of identifier info, then sends the same 20 bytes back
+            byte[] header = packet.Read(20);
```
AlanMorel, line 31: I think a better name than `header` should be used here since header is usually referring to the opcode

MapleServer2/Packets/JobPacket.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            // Ordered list of skill ids (must be sent in this order)
+            int[] ids = JobSkillOrder.Order[character.JobGroupId];
+            int countId = ids[ids.Length - 8]; // 8th to last skill id
+
+            pWriter.WriteByte((byte)(ids.Length - 8)); // Skill count minus 8
+
+            // List of skills for given tab in format (byte zero) (byte learned) (int skill_id) (int skill_level) (byte zero)
+            foreach (int id in ids) {
+                if (id == countId) pWriter.WriteByte(8); // Write that there are 8 skills left
```
AlanMorel, line 48: Can braces be added here? I think it improves readability

MapleServer2/Packets/SkillBookTreePacket.cs
```diff
@@ -0,0 +1,43 @@
+using System.Collections.Generic;
+using MaplePacketLib2.Tools;
+using MapleServer2.Constants;
+using MapleServer2.Types;
+
+namespace MapleServer2.Packets {
+    public static class SkillBookTreePacket {
+        private static int count = 0;
```
AlanMorel, line 8: What is this variable doing?

MapleServer2/Types/Player.cs
```diff
@@ -87,19 +90,42 @@ public class Player {
             stats.MoveSpd = new PlayerStat(110, 100, 150);
             stats.JumpHeight = new PlayerStat(110, 100, 130);
 
-            return new Player
-            {
+            List<SkillTab> skillTabs = new List<SkillTab>();
+            skillTabs.Add(new SkillTab("test1"));
+
+            // Add runeblade skills set to level 1 and unlearned (normally would be loaded from database)
+            XmlDocument xmlDoc= new XmlDocument();
```
AlanMorel, line 97: Anything dealing with XML should definitely happen in another class

MapleServer2/Types/SkillTab.cs
```diff
@@ -1,74 +1,54 @@
...
-                }
-            }
+        public void AddOrUpdate(Skill skill) {
+            // Add or update skill
+            if (Skills.ContainsKey(skill.Id)) Skills[skill.Id] = skill;
+            else Skills.Add(skill.Id, skill);
 
-            skills.Add(skill);
+            // Recursive add or update for sub skills
+            if (skill.Sub != null) foreach (int sub in skill.Sub) AddOrUpdate(Skill.skill(sub, skill.Level, skill.Learned, skill.Feature));
```
AlanMorel, line 27: Can braces be added to this entire method?

MapleServer2/Types/Skill.cs
```diff
@@ -0,0 +1,21 @@
+namespace MapleServer2.Types {
+    public struct Skill {
+        public int Id { get; private set; }
+        public short Level { get; private set; }
+        public byte Learned { get; private set; }
+        public string Feature { get; private set; }
+        public int[] Sub { get; private set; }
+
+        public static Skill skill(int id, short level, byte learned, string feature = "", int[] sub = null) {
```
AlanMorel, line 9: Does this method need to exist if it's just calling the constructor and passing in everything 1 to 1? Can't the constructor just be called directly instead?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/JobHandler.cs
Sparkymod: In this case byte is because in each count, the packet will read that last byte and move the offset to the next Skill info structure so abool will return a true or false statement and he wants to see if is 00 or 01 but also move to the next one to keep reading.

MapleServer2/Constants/Skills/JobSkillOrder.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace MapleServer2.Constants.Skills {
+    public static class JobSkillOrder {
+        public static Dictionary<int, int[]> Order = new Dictionary<int, int[]>();
+
+        static JobSkillOrder() {
+            Order.Add(10, new int[] {
+                10100091, 10100261, 10100041, 10100211, 10100161, 10100281, 20000011, 10100061, 10100011, 10100181,
+                10100062, 10100131, 10100081, 10100031, 10100201, 10100151, 10100101, 20000001, 10100051, 10100221,
+                10100001, 10100171, 10100291, 10100071, 10100021, 10100191, 10100141, 10100294, 10100111, 10100162,
+                10100231, 10100251, 10100202, 10100271, 10100204, 10100121, 10100241
```
newgithubaccountlol, line 12: Any way we can move these into either a db that handles these constants or a file?

MapleServer2/Packets/SkillBookTreePacket.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static Packet Open(Player character) {
+            var pWriter = PacketWriter.Of(SendOp.SKILL_BOOK_TREE);
+
+            // Writes only skills that are learned and for the job rank tab that is opened also doesn't write default passive skills
+            pWriter.WriteByte(0); // Mode (0 = open) (1 = save)
```
newgithubaccountlol, line 13: There's a few other modes (like 2 and 4).
2 is long, unicodestring, byte
4 is int
Just putting this here for more info

MapleServer2/Packets/SkillBookTreePacket.cs
```diff
@@ -0,0 +1,43 @@
...
+            pWriter.WriteInt(1); // Possibly always 1
+            pWriter.WriteLong(character.SkillTabs[0].Id); // Some identifier (skill tab id?)
+            pWriter.WriteInt(1); // Repeat of last int and long
+            pWriter.WriteLong(character.SkillTabs[0].Id); // Same as previous identifier (skill tab id?)
+            pWriter.WriteUnicodeString("Build 1"); // Build name?
```
newgithubaccountlol, line 18: Page name

MapleServer2/Packets/SkillBookTreePacket.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            return pWriter;
+        }
+
+        public static Packet Save(byte[] header) {
+            var pWriter = PacketWriter.Of(SendOp.SKILL_BOOK_TREE);
+
+            pWriter.WriteByte(1); // Mode (0 = open) (1 = save)
+            pWriter.Write(header); // Echoes first 20 bytes of REQUEST_SKILL_BOOK_TREE, probably some identifier
```
newgithubaccountlol, line 38: 2 longs of the serial, and then an int

MapleServer2/Types/Player.cs
newgithubaccountlol: Agree with this. Players have nothing to do with XML

MapleServer2/Types/Player.cs
```diff
@@ -129,8 +157,31 @@ public static Player MaleDefault(long accountId, long characterId, string name =
...
             stats.JumpHeight = new PlayerStat(110, 100, 130);
 
-            return new Player
-            {
+            List<SkillTab> skillTabs = new List<SkillTab>();
+            skillTabs.Add(new SkillTab("test1"));
+
+            // Add archer skills set to level 1 and unlearned (normally would be loaded from database)
+            XmlDocument xmlDoc= new XmlDocument();
+            xmlDoc.Load("./MapleServer2/Constants/Skills/skills_50.xml");
```
newgithubaccountlol, line 165: This code looks duplicated below, maybe put it into one method that loads skills based on job (in this case 50)

MapleServer2/Types/Skill.cs
newgithubaccountlol: To add on to the previous comment, if done this way, these objects should be cached somewhere because that's the general usage of static as some sort of factory. It's not immediately obvious how this is used but from the example in how a new character is made and how job skills are loaded later, it looks it this is a factory of some sort so it should be cached.

newgithubaccountlol (commented):
Generally good, not a huge fan of the XML in player class but other than that, looks pretty fine. Maybe revisit the skill ctor

MapleServer2/Types/Player.cs
Sparkymod: agree to this

MapleServer2/PacketHandlers/Game/JobHandler.cs
AlanMorel: Again, if there are only two states, a boolean makes sense here. Also the name `learned` implies you can only ever have two states, you either learned the skill or you didn't. If there are more valid values beyond `00` and `01`, then I agree that a byte is better, but not if the only values are `00` and `01`

davidshi-teal: That is not how the client works though. I think there is some missing knowledge here. 

The client basically sends this packet to the server and the server is handling what the client sent. You can read this particular byte as 50 always and the client will still send over the proper information the next time because how the server handles the code has nothing to do with the client (assuming you don't send over bad data back to the client). You can test this by removing all code that the server sends back to the client and see what the values are here. It should stay as 0 or 1 depending on the option. What you are saying is that the client is stateful based on what the server read which doesn't really make much sense.

The rest of the understanding is also incorrect. The reason he picked byte is probably because that byte _is_ a byte and not another data type. It's pretty fair to save a byte that's sent over by the client in a byte variable. What sharpacex is saying is that because the byte only has 2 states, it should be a boolean.

However, I don't feel like it's necessary to change this to a boolean because we don't know **for sure** if there are more states.

AlanMorel: Then there is still an inconsistency here, except now it's in the name. If we are going off the assumption that there could be more states than 2, then `learned` is not a good name for this. What does `learned` being equal to `5` mean then? Maybe `learnState` or something that implies "hey we think this has two states, but it could have more". I see `learned` and immediately think okay so you either learned it or not. Either way, I don't feel that strongly about this, but I do think it is still inconsistent to use a name that implies only 2 states, but give it a data type that implies more than 2 states.

MapleServer2/Types/Player.cs
rhydev: Those are hard-coded players only there for testing, so I don't think it's necessary here.

MapleServer2/Constants/Skills/JobSkillOrder.cs
eotetb: I'm going to second this. We should really create db files for this type of thing rather than baking it into the code.

rhydev: Could move them into an xml or something but don't really see the point of doing that just to parse it and end up with the same objects.

MapleServer2/Data/Static/ItemMetadataStorage.cs
```diff
@@ -1,15 +1,15 @@
...
+using Maple2Storage.Types;
 
 namespace MapleServer2.Data.Static {
     // This is an in-memory storage to help with determining some metadata of items
     public static class ItemMetadataStorage {
         private static readonly Dictionary<int, ItemMetadata> map = new Dictionary<int, ItemMetadata>();
 
         static ItemMetadataStorage() {
-            using FileStream stream = File.OpenRead("Resources/ms2-item-metadata");
+            using FileStream stream = File.OpenRead("Maple2Storage/Resources/ms2-item-metadata");
```
eotetb, line 12: Is there a good reason for changing the filepath?

MapleServer2/Types/Skill.cs
rhydev: Tbh I copied the structure and then changed it from another type in Maple2.Data iirc and yeah it can just use the constructor.

MapleServer2/Types/SkillTab.cs
rhydev: Agreed braces are better here.

MapleServer2/Types/Player.cs
rhydev: Disagree to this, unless there is an overarching XML parsing with more XML data, because this would only ever be parsed when creating a new character, so there is no reason to move it out of this class until it would be used more than 1 time.

MapleServer2/Packets/SkillBookTreePacket.cs
rhydev: Left over variable from reverse engineering that I forgot to remove

MapleServer2/Packets/JobPacket.cs
rhydev: Personally think that it reads better without braces since then it's all single line within the for loop.

MapleServer2/PacketHandlers/Game/SkillBookTreeHandler.cs
rhydev: Agreed that header is a poor name and thought about that when I named it, open to suggestions for better name.

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
```diff
@@ -61,8 +61,8 @@ public void HandleCreate(LoginSession session, PacketReader packet)
         {
             byte gender = packet.ReadByte();
             //packet.ReadShort(); // const?
-            var jobCode = (Job)packet.ReadShort();
-            //int jobCode = packet.ReadShort();
+            // var jobCode = (Job)packet.ReadShort();
```
eotetb, line 64: is there a reason for keeping the commented out code?

MapleServer2/PacketHandlers/Game/JobHandler.cs
rhydev: I agree with both sides here, it's a matter of deciding how we want to store data, personally, I think it should use the same type as what is being read. For instance, skill level is sometimes read as a short and sometimes as an int, but in both cases it only takes up the data of a byte. I prefer to keep skill level as a short as that is what is most commonly read, and since the data type is built off of read data, I think it should match the most common data type.

MapleServer2/PacketHandlers/Game/JobHandler.cs
rhydev: Agreed.

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
```diff
@@ -61,8 +61,8 @@ public void HandleCreate(LoginSession session, PacketReader packet)
...
+            int jobCode = packet.ReadShort();
             string name = packet.ReadUnicodeString();
             var skinColor = packet.Read<SkinColor>();
```
eotetb, line 67: Since we changed jobCode to use an explicit datatype, we should make sure the other variables are also using explicit datatypes to be consistent.

MapleServer2/Constants/RecvOp.cs
rhydev: Weird artifact from merging, I actually changed that to uppercase X but for some reason git decided that I did the opposite.

MapleServer2/Packets/FieldPacket.cs
```diff
@@ -124,7 +123,7 @@ public static class FieldPacket {
                 pWriter.WriteInt();
                 pWriter.WriteShort();
             } else {
-                pWriter.WriteInt();
+                //pWriter.WriteInt(); commented out to remove warning
```
eotetb, line 126: Are we currently aware as to why this is throwing a warning? I typically don't recommend commenting stuff out just to hide warnings because we will forget about it.

MapleServer2/Packets/JobPacket.cs
```diff
@@ -0,0 +1,61 @@
...
+            return pWriter;
+        }
+
+        public static Packet WriteSkills(PacketWriter pWriter, Player character) {
+            // Get skills
+            Dictionary<int, Skill> skills = character.SkillTabs[0].GetSkills(); // Get first skill tab skills only for now, uncertain of how to have multiple skill tabs
+
+            // Ordered list of skill ids (must be sent in this order)
+            int[] ids = JobSkillOrder.Order[character.JobGroupId];
+            int countId = ids[ids.Length - 8]; // 8th to last skill id
```
eotetb, line 42: Do we know what the 8 offset is for? If we need it, I would create another variable for the offset since this is reused multiple times.

MapleServer2/Types/Player.cs
eotetb: A class should only contain functionality that pertains to the objects it would create. As said above, XML should be its own separate class. We can call the method from here still if needed, but it should be broken out.

MapleServer2/Types/SkillTab.cs
```diff
@@ -1,74 +1,54 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.InteropServices;
-using MaplePacketLib2.Tools;
 
-namespace Maple2.Data.Types {
-    public class SkillTab : IByteSerializable {
+namespace MapleServer2.Types {
```
eotetb, line 4: Is there any reason we changed the namespace?

MapleServer2/Packets/SkillBookTreePacket.cs
rhydev: It's known exactly what the data is, it's the same as the data in saving skill tree, could update it but the data is redundant and doesn't do anything, just requires the first 20 bytes only.

MapleServer2/Packets/SkillBookTreePacket.cs
rhydev: Can you provide any use cases or sniffs for the other modes? I haven't seen those modes used.

MapleServer2/Types/SkillTab.cs
```diff
@@ -1,74 +1,54 @@
...
-            // Iterate backwards for safe modification
-            for (int i = skills.Count - 1; i >= 0; i--) {
-                if (skills[i].SkillId == skill.SkillId) {
-                    skills.RemoveAt(i);
-                }
-            }
+        public void AddOrUpdate(Skill skill) {
+            // Add or update skill
+            if (Skills.ContainsKey(skill.Id)) Skills[skill.Id] = skill;
+            else Skills.Add(skill.Id, skill);
```
eotetb, line 24: While to be fair, we haven't decided on code standards yet, having everything on a single line can make the code a bit harder to read at a quick glance, which is important when you're looking for defects.

eotetb (commented):
I recommend in the future to try to create smaller but more frequent commits and pull requests. This is a lot of changes to read through in a single pull request, and I feel that not all of it is related to the skill tree.

MapleServer2/Constants/Skills/JobSkillOrder.cs
eotetb: The reason for moving it out is mainly for organization, we would handle similar functions in a similar manner. Better organization improves maintainability, even if it provides no performance gains.

MapleServer2/Packets/FieldPacket.cs
rhydev: Unreachable code from a hardcoded if statement because the packet isn't entirely decrypted.

MapleServer2/Packets/JobPacket.cs
rhydev: It can be almost any number, I've seen it be 8, 9, 10, 11, 12, 13, 14, and 15, but it all works the same with 8, not sure why it is written this way to be honest and I don't think it's used in other packets.

MapleServer2/Packets/FieldPacket.cs
Sparkymod: This one is because it never reach the _else_ as it said the warning.

MapleServer2/Types/SkillTab.cs
rhydev: Yes, that path no longer exists, the file is under MapleServer2/Types, it was previously copy pasted from a different directory without changing namespace before I touched the file.

MapleServer2/Types/SkillTab.cs
rhydev: Agreed that formatting is relative, I prefer one line ifs and elses without brackets, but if it is decided that all ifs and elses should have brackets it can be changed.

MapleServer2/Data/Static/ItemMetadataStorage.cs
rhydev: Yes, before the working directory was Maple2Storage, this change changes the working directory to the solution directory.

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
rhydev: Since I don't know the origins of why the previous code was created I left it commented out.

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
rhydev: Which variables are you referring to here?

Sparkymod (approved):
Other than the XML change, all should be good then

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
eotetb: Late response I know, but we have skinColor for example declared implicitly (var), while other variables here are declared explicitly.

PR #4, opened 2020-12-09
Feat: skill tree (review)

rhydev:
- Updated formatting to new project standards for relevant files
- Removed JobSkillOrder class and added order attribute to skill xml files
- Added order property to SkillTab
- Added XmlParser class in Tools to parse skill xmls
- Fixed Skill class to use constructor instead of method
- Updated SkillBookTreePacket.Save to load actual data instead of just parroting client request
- Fixed various small comments


Sparkymod:
Is all good, fixing the last PR issues

newgithubaccountlol (approved):

PR #5, opened 2020-12-09
Cleaning up & formatting

Sparkymod:
- Removing some files and folder that are not need it.
- Cleaning up excess of line and formatting the project to the new standards.


MapleServer2/PacketHandlers/Game/StateSkillHandler.cs
```diff
@@ -3,25 +3,29 @@
...
-        public override void Handle(GameSession session, PacketReader packet) {
+        public override void Handle(GameSession session, PacketReader packet)
+        {
             byte function = packet.ReadByte();
-            if (function == 0) {
+            if (function == 0)
+            {
                 // This count seems to increase for each skill used
-                int counter = packet.ReadInt();
+                packet.ReadInt();   // Counter
```
AlanMorel, line 20: Why remove the variable declaration just to add the variable name as a comment next to it?

MapleServer2/Types/FieldState.cs
```diff
@@ -1,72 +1,76 @@
...
+        public void AddPortal(IFieldObject<Portal> portal)
+        {
             portals[portal.ObjectId] = portal;
         }
 
-        public bool RemovePortal(int objectId) {
-            return portals.Remove(objectId, out _);
-        }
-
-        public void AddMob(IFieldObject<Mob> mob)
```
AlanMorel, line 67: Why was this removed?

MapleServer2/Types/Mob.cs
```diff
@@ -1,19 +0,0 @@
...
-        public byte Animation;
-        public PlayerStats stats;
-
-        public Mob(int id)
-        {
-            this.Id = id;
-            this.Animation = 255;
-        }
-    }
-}
```
AlanMorel, line 19: I do not understand why we are deleting things that we will eventually need, this makes zero sense.

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -5,21 +5,25 @@
...
         public UserChatHandler(ILogger<GamePacketHandler> logger) : base(logger) { }
 
-        public override void Handle(GameSession session, PacketReader packet) {
-            var type = (ChatType) packet.ReadInt();
+        public override void Handle(GameSession session, PacketReader packet)
+        {
+            ChatType type = (ChatType)packet.ReadInt();
             string message = packet.ReadUnicodeString();
-            string recipient = packet.ReadUnicodeString();
+            packet.ReadUnicodeString(); // Recipient
```
AlanMorel, line 20: Again why remove the variable

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/StateSkillHandler.cs
Sparkymod: Removing variable that is not being used, but add the comment to keep track for what it was using it in further implementations.

MapleServer2/Types/FieldState.cs
Sparkymod: The "Mob" class and all his implementation, was something i made for testing before, to see if his behavior was the same as NPC, so removing it from the source would be better to not confuse people that was implemented.

MapleServer2/Types/Mob.cs
Sparkymod: Again, the mob was made for testing purpose, and is not sure that is the same as NPC

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
Sparkymod: The variable is not being used, so for not lose track, i added as comment to remember what it was

MapleServer2/Types/Mob.cs
AlanMorel: When I implemented mob spawning on my end, I used this existing implementation as the base, so this is code worth keeping and later improving once the emulator gets there

newgithubaccountlol:
I don't have any big comments for this PR but I would probably do 1 formatting PR (only formatting) and then another PR for changes like cleanup. With a huge change like this I don't even want to look at the cleanup changes

AlanMorel:
> I don't have any big comments for this PR but I would probably do 1 formatting PR (only formatting) and then another PR for changes like cleanup. With a huge change like this I don't even want to look at the cleanup changes

Completely agree, I still went through every file but formatting is just noise and should have been done separately from clean up so the clean up can actually be reviewable on its own

PR #9, opened 2020-12-13
Implemented Item Stacking

bwu113:
Added/Modified code to implement item stacking.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
@@ -1,17 +1,22 @@
-﻿using MaplePacketLib2.Tools;
+﻿using System;
+using System.Diagnostics;
+using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
 using MapleServer2.Packets;
 using MapleServer2.Servers.Game;
 using MapleServer2.Types;
 using Microsoft.Extensions.Logging;
 
-namespace MapleServer2.PacketHandlers.Game {
-    public class FieldEnterHandler : GamePacketHandler {
+namespace MapleServer2.PacketHandlers.Game
+{
+    public class FieldEnterHandler : GamePacketHandler
+    {
         public override RecvOp OpCode => RecvOp.RESPONSE_FIELD_ENTER;
 
         public FieldEnterHandler(ILogger<FieldEnterHandler> logger) : base(logger) { }
 
-        public override void Handle(GameSession session, PacketReader packet) {
+        public override void Handle(GameSession session, PacketReader packet)
+        {
             packet.ReadInt(); // ?
 
             // Liftable: 00 00 00 00 00
@@ -27,7 +32,8 @@ namespace MapleServer2.PacketHandlers.Game {
             QuickSlot eagleGlide = QuickSlot.From(10500151);
             QuickSlot testSkill = QuickSlot.From(10500153);
 
-            if (session.Player.GameOptions.TryGetHotbar(0, out Hotbar mainHotbar)) {
+            if (session.Player.GameOptions.TryGetHotbar(0, out Hotbar mainHotbar))
+            {
                 /*
                 mainHotbar.MoveQuickSlot(4, arrowStream);
                 mainHotbar.MoveQuickSlot(5, arrowBarrage);
@@ -38,30 +44,39 @@ namespace MapleServer2.PacketHandlers.Game {
             }
 
             // Add catalysts for testing
-            
-            int[] catalysts = { 40100001, 40100002, 40100003, 40100021, 40100023, 40100024, 40100026 };
+
+            int[] catalysts = { 40100001, 40100002, 40100003, 40100021, 40100023, 40100024, 40100026, 15100216, 15100216 };
 
             // Adds first set of item test case
-            foreach(int catalyst in catalysts)
+
+            Item item = new Item(40100001)
             {
-                Item item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+                Amount = 10,
+                Uid = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0)
+            };
+            Item item2 = new Item(40100001)
+            {
+                Amount = 8,
+                Uid = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0)
+            };
+            Item item3 = new Item(40100001)
+            {
+                Amount = 5,
+                Uid = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0)
+            };
 
-                session.Inventory.Add(item);
-                session.Send(ItemInventoryPacket.Add(item));
-            }
+            session.Inventory.Add(item);
+            session.Inventory.Add(item2);
+            session.Inventory.Add(item3);
 
-            // Adds 2nd set of same items to test stacking
-            foreach (int catalyst in catalysts) {
-                Item item = new Item(catalyst) { Amount = 4, Uid = catalyst };
-                bool doesExist = session.Inventory.Add(item);
-                if (!doesExist) {
-                    session.Send(ItemInventoryPacket.Add(item));
-                }
-                else
-                {
-                    session.Inventory.Update(item, (item.Amount + item.Amount));
-                    session.Send(ItemInventoryPacket.Update(item.Uid, (item.Amount + item.Amount)));
-                }
+            Debug.WriteLine(item.Amount);
+
+            session.Send(ItemInventoryPacket.Add(item));
+            if (item2.Amount > 0)
+            {
+                Debug.WriteLine(item2.Amount);
+                session.Send(ItemInventoryPacket.Add(item2));
+                session.Send(ItemInventoryPacket.Add(item3));
             }
         }
     }
--- a/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
@@ -38,14 +38,32 @@ namespace MapleServer2.PacketHandlers.Game {
             }
 
             // Add catalysts for testing
-            /*
+            
             int[] catalysts = { 40100001, 40100002, 40100003, 40100021, 40100023, 40100024, 40100026 };
-            foreach (int catalyst in catalysts) {
-                var item = new Item(catalyst) { Amount = 99999, Uid = catalyst };
+
+            //Adds first set of item test case
+            foreach(int catalyst in catalysts)
+            {
+                var item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+
                 session.Inventory.Add(item);
                 session.Send(ItemInventoryPacket.Add(item));
             }
-            */
+
+            //Adds 2nd set of same items to test stacking
+            foreach (int catalyst in catalysts) {
+                var item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+
+                if (!session.Inventory.Add(item))
+                {
+                    session.Send(ItemInventoryPacket.Update(item.Uid, (item.Amount + item.Amount)));
+                }
+                else
+                {
+                    session.Send(ItemInventoryPacket.Add(item));
+                }
+            }
+            
         }
     }
 }
--- a/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
@@ -70,11 +70,9 @@ namespace MapleServer2.PacketHandlers.Game
             session.Inventory.Add(item3);
 
             session.Send(ItemInventoryPacket.Add(item));
-            if (item2.Amount > 0)
-            {
-                session.Send(ItemInventoryPacket.Add(item2));
-                session.Send(ItemInventoryPacket.Add(item3));
-            }
+            session.Send(ItemInventoryPacket.Add(item2));
+            session.Send(ItemInventoryPacket.Add(item3));
+            
         }
     }
 }
--- a/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
@@ -49,6 +49,7 @@ namespace MapleServer2.PacketHandlers.Game {
                 session.Inventory.Add(item);
                 session.Send(ItemInventoryPacket.Add(item));
             }
+            //session.Inventory.print();
 
             //Adds 2nd set of same items to test stacking
             foreach (int catalyst in catalysts) {
@@ -56,14 +57,15 @@ namespace MapleServer2.PacketHandlers.Game {
 
                 if (!session.Inventory.Add(item))
                 {
+                    session.Inventory.Update(item, (item.Amount + item.Amount));
                     session.Send(ItemInventoryPacket.Update(item.Uid, (item.Amount + item.Amount)));
                 }
                 else
                 {
                     session.Send(ItemInventoryPacket.Add(item));
                 }
+            //session.Inventory.print();
             }
-            
         }
     }
 }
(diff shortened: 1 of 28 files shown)


MapleServer2/Types/Inventory.cs
```diff
@@ -45,6 +45,11 @@ public class Inventory {
             // Item has a slot set, try to use that slot
             if (item.Slot >= 0) {
                 if (!SlotTaken(item, item.Slot)) {
+                    //Performs a item exist check before adding item
+                    if (Check(item))
```
AlanMorel, line 49: Does it make sense to add this `Check` method directly inside AddInternal since it seems like you'd always want to do this check before calling it?

bwu113:
```suggestion
                   if (!session.Inventory.Add(item))
                {
                    session.Inventory.Update(item, (item.Amount + item.Amount));
                    session.Send(ItemInventoryPacket.Update(item.Uid, (item.Amount + item.Amount)));
                }
                else
                {
                    session.Send(ItemInventoryPacket.Add(item));
                }
```
Issue is that if I bring the `Check` method directly inside AddInternal, There is no way for the above code in `FieldEnterHandler` to know whether to run `ItemInventoryPacket.Add` or `ItemInventoryPacket.Update`

Feel free to experiment with that. I'm gonna mess with it again later tonight.

MapleServer2/Types/Inventory.cs
```diff
@@ -132,15 +152,38 @@ public class Inventory {
             }
         }
 
+        // Checks if item exists already or if slot is taken already.
+        public bool Check(Item item)
+        {
+            if (items.ContainsKey(item.Uid)){
+                return true;
+            }
+            if (GetSlots(item.InventoryType).ContainsKey(item.Slot))
```
newgithubaccountlol, line 161: Optional at this point but the items have max slot sizes, should we check this at this point?

bwu113: Probably. Will have to determine max slot size for items.

MapleServer2/Types/Inventory.cs
```diff
@@ -132,15 +152,38 @@ public class Inventory {
...
+        }
+
         // This REQUIRES item.Slot to be set appropriately
         private void AddInternal(Item item) {
-            Debug.Assert(!items.ContainsKey(item.Uid),
-                "Error adding an item that already exists");
-            items[item.Uid] = item;
+            /*Debug.Assert(!items.ContainsKey(item.Uid),
+                "Error adding an item that already exists");*/
+                items[item.Uid] = item;
```
eotetb, line 182: While I'm not super concerned about some of the formatting since we haven't decided on coding standards, we want to try to keep code as clean as possible. You generally only want your code to be tabbed over by one in whatever block its in. The way it was formatted before was fine.

MapleServer2/Types/Inventory.cs
bwu113: can easily add that to the `Check` Method have it ran in the `Add` Method.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -38,14 +38,34 @@ public class FieldEnterHandler : GamePacketHandler {
...
+            //Adds first set of item test case
+            foreach(int catalyst in catalysts)
+            {
+                var item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+
                 session.Inventory.Add(item);
                 session.Send(ItemInventoryPacket.Add(item));
             }
-            */
+            //session.Inventory.print();
```
eotetb, line 52: We want to get rid of test code like this after we're done testing with it, we don't want to leave commented out test code in the master branch.

eotetb (commented):
I see that there is some test code within the source. It might be worth writing unit tests for this kind of thing in the future. It does take a bit more time to setup, but we can keep running them afterwards to make sure changes don't change expected behavior.

Aside from stray test code or commented out test code, and some formatting issues, the actual changes seem to look good to me.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
bwu113: Gotcha will update.

bwu113:
Removed all Test code.

MapleServer2/Types/Inventory.cs
rhydev: This is not resolved, please untab this and the below GetSlot().

MapleServer2/Types/Inventory.cs
```diff
@@ -45,6 +45,11 @@ public class Inventory {
             // Item has a slot set, try to use that slot
             if (item.Slot >= 0) {
                 if (!SlotTaken(item, item.Slot)) {
+                    //Performs a item exist check before adding item
```
rhydev, line 48: Comments should have a space after `//`

MapleServer2/Types/Inventory.cs
```diff
@@ -132,15 +152,28 @@ public class Inventory {
             }
         }
 
+        // Checks if item exists already or if slot is taken already.
+        public bool Check(Item item)
```
rhydev, line 156: Should be renamed to `Exists` since it returns a boolean.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -38,14 +38,32 @@ public class FieldEnterHandler : GamePacketHandler {
...
                 session.Inventory.Add(item);
                 session.Send(ItemInventoryPacket.Add(item));
             }
-            */
+
+            //Adds 2nd set of same items to test stacking
+            foreach (int catalyst in catalysts) {
+                var item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+
+                if (!session.Inventory.Add(item))
```
rhydev, line 57: Generally bad practice to call methods inside ifs that mutate data, also if you're going to have an if else, generally you should handle the truthy statement before the false one.
```suggestion
                bool added = session.Inventory.Add(item));
                if (!added)
```

rhydev (commented):
Please use `Shift + Alt + F` to format files. We probably need to have some style standard at least in the readme.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -41,27 +41,26 @@ public class FieldEnterHandler : GamePacketHandler {
...
+            // Adds 2nd set of same items to test stacking
             foreach (int catalyst in catalysts) {
-                var item = new Item(catalyst) { Amount = 4, Uid = catalyst };
-
-                if (!session.Inventory.Add(item))
-                {
-                    session.Inventory.Update(item, (item.Amount + item.Amount));
-                    session.Send(ItemInventoryPacket.Update(item.Uid, (item.Amount + item.Amount)));
+                Item item = new Item(catalyst) { Amount = 4, Uid = catalyst };
+                bool doesExist = session.Inventory.Add(item);
```
rhydev, line 56: Can just name `exists` but this is ok.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -41,27 +41,26 @@ public class FieldEnterHandler : GamePacketHandler {
...
+                if (!doesExist) {
```
rhydev, line 57: Should be (doesExist) not (!doesExist) since you flipped the inner code.

MapleServer2/Types/Inventory.cs
```diff
@@ -45,10 +45,10 @@ public class Inventory {
             // Item has a slot set, try to use that slot
             if (item.Slot >= 0) {
                 if (!SlotTaken(item, item.Slot)) {
-                    //Performs a item exist check before adding item
-                    if (Check(item))
+                    // Performs a item exist check before adding item
+                    if (Exists(item))
                     {
-                        return false;
+                        return Exists(item);
```
rhydev, line 51: This is just repetitive and waste of processing, revert to return false or add a variable above the if with the value of `Exists(item)`

MapleServer2/Types/Inventory.cs
```diff
@@ -60,10 +60,10 @@ public class Inventory {
             for (short i = 0; i < Size; i++) {
                 if (!SlotTaken(item, i)) {
                     item.Slot = i;
-                    //Performs a item exist check before adding item
-                    if (Check(item))
+                    // Performs a item exist check before adding item
+                    if (Exists(item))
                     {
-                        return false;
+                        return Exists(item);
```
rhydev, line 66: Same as above

rhydev (commented):
Still needs to fix brace styling

MapleServer2/Types/Inventory.cs
dilsonnn: I would extract it into a method addItem and reuse since the code fragments are basically the same

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,134 @@
...
+                    }
                     return true;
                 }
             }
 
             return false;
         }
 
+        // Checks if item is stackable.
+        private bool isStackable(Item item) => item.SlotMax > 1;
```
rhydev, line 97:
```suggestion
        private bool isStackable(Item item)
        {
            return item.SlotMax > 1;
        }
```
Please use brackets.

MapleServer2/Types/Inventory.cs
```diff
@@ -83,9 +144,12 @@ public class Inventory {
             return item.TrySplit(amount, out removedItem) ? item.Amount : -1;
         }
 
+
```
rhydev, line 147: Why is there an extra new line now?

MapleServer2/Types/Inventory.cs
```diff
@@ -118,22 +185,26 @@ public class Inventory {
...
             slots.Clear();
-            for (short i = 0; i < tabItems.Count; i++) {
+            for (short i = 0; i < tabItems.Count; i++)
+            {
                 tabItems[i].Slot = i;
                 slots[i] = tabItems[i].Uid;
             }
         }
 
+
```
rhydev, line 204: Extra new line again

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,134 @@
...
-            for (short i = 0; i < Size; i++) {
-                if (!SlotTaken(item, i)) {
+            for (short i = 0; i < Size; i++)
+            {
+                if (!SlotTaken(item, i))
+                {
                     item.Slot = i;
-                    AddInternal(item);
+
+
```
rhydev, line 80: Why 2 new lines here?

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,134 @@
...
+                        else
+                        {
+                            i.Amount = i.Amount + item.Amount;
+                            return;
+                        }
+                    }
+                }
+            }
+            AddInternal(item);
+        }
```
rhydev, line 121: Still using test value of 10?

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,134 @@
...
+namespace MapleServer2.Types
+{
+    public class Inventory
+    {
         public short Size { get; }
         public IReadOnlyDictionary<long, Item> Items => items;
-
+        int originAmount;
+        int amount;
+        bool isStacking;
```
rhydev, line 17: These properties are unused.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -38,14 +44,40 @@ public class FieldEnterHandler : GamePacketHandler {
...
+            {
+                Amount = 5,
+                Uid = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0)
+            };
+
+            session.Inventory.Add(item);
+            session.Inventory.Add(item2);
+            session.Inventory.Add(item3);
+
+            Debug.WriteLine(item.Amount);
```
rhydev, line 72: Remove console logging.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -38,14 +44,40 @@ public class FieldEnterHandler : GamePacketHandler {
...
+
+            session.Send(ItemInventoryPacket.Add(item));
+            if (item2.Amount > 0)
+            {
+                Debug.WriteLine(item2.Amount);
```
rhydev, line 77: Remove console logging.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -38,14 +44,40 @@ public class FieldEnterHandler : GamePacketHandler {
...
+                session.Send(ItemInventoryPacket.Add(item2));
+                session.Send(ItemInventoryPacket.Add(item3));
             }
```
rhydev, line 80: Doesn't make sense that `item3` is only added if `item2.Amount > 0`. Also an `InventoryController` of sorts should determine which inventory packets should be sent.

rhydev (changes requested):
Still needs some work but has improved.

MapleServer2/Types/Inventory.cs
bwu113: ?? i didn't add that. its always been there

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
bwu113: I just added item 3 to test if it would stack correctly in a subsequent stacking situation. 

IE. 1st item is stacked to full, 2nd item is stacked to full. 3rd item is left over.

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,131 @@
...
+        private bool isStackable(Item item)
+        {
+            return item.SlotMax > 1;
+        }
+
+        private void Stack(Item item) // Replace 10 with item.slotmax to test against actual max slot count.
+        {
+            foreach (Item i in items.Values)
+            {
+                if (i.Id == item.Id)
```
AlanMorel, line 100: Can you invert this condition and if this new condition is true, add a `continue;`?

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,131 @@
...
+                {
+                    if (i.Amount < i.SlotMax)
```
AlanMorel, line 102: Same as above

AlanMorel (changes requested):

MapleServer2/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "MapleServer2": {
+      "commandName": "Project",
+      "workingDirectory": "E:\\MS2 Pserver\\MapleStory 2 Server Emulator"
```
integralcoder95: Can you change the working directory to `$(SolutionDir)`
It should fix the path problem. Or add this file to gitignore in the future :D

rhydev: This file isn't included in the PR where did this come from?

bwu113: This file is on my end. thats why its deleted off this PR

MapleServer2/Types/Inventory.cs
```diff
@@ -6,73 +6,130 @@
...
                 }
 
                 item.Slot = -1; // Reset slot
             }
 
-            for (short i = 0; i < Size; i++) {
-                if (!SlotTaken(item, i)) {
+            for (short i = 0; i < Size; i++)
+            {
+                if (!SlotTaken(item, i))
```
AlanMorel, line 73: Could flip condition and use continue here as well

AlanMorel (changes requested):

PR #10, opened 2020-12-14
Adding type of chat

Sparkymod:
- Super(hardcoded item), World and Channel.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
@@ -15,7 +15,7 @@ namespace MapleServer2.PacketHandlers.Game
 
         public override void Handle(GameSession session, PacketReader packet)
         {
-            var type = (ChatType)packet.ReadInt();
+            ChatType type = (ChatType)packet.ReadInt();
             string message = packet.ReadUnicodeString();
             string recipient = packet.ReadUnicodeString();
             packet.ReadLong();
--- a/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
@@ -21,21 +21,7 @@ namespace MapleServer2.PacketHandlers.Game
             packet.ReadLong();
 
             GameCommandActions.Process(session, message);
-            switch (type)
-            {
-                case (ChatType.Super):
-                    session.FieldManager.SendChat(session.Player, message, type);
-                    break;
-                case (ChatType.World):
-                    session.FieldManager.SendChat(session.Player, message, type);
-                    break;
-                case (ChatType.Channel):
-                    session.FieldManager.SendChat(session.Player, message, type);
-                    break;
-                default:
-                    session.FieldManager.SendChat(session.Player, message);
-                    break;
-            }
+            session.FieldManager.SendChat(session.Player, message, type);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -5,22 +5,36 @@
...
+    {
         public override RecvOp OpCode => RecvOp.USER_CHAT;
 
         public UserChatHandler(ILogger<GamePacketHandler> logger) : base(logger) { }
 
-        public override void Handle(GameSession session, PacketReader packet) {
-            var type = (ChatType) packet.ReadInt();
+        public override void Handle(GameSession session, PacketReader packet)
+        {
+            var type = (ChatType)packet.ReadInt();
```
AlanMorel, line 18: Space after the cast

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -5,22 +5,36 @@
...
+                    break;
+                case (ChatType.World):
+                    session.FieldManager.SendChat(session.Player, message, type);
+                    break;
+                case (ChatType.Channel):
+                    session.FieldManager.SendChat(session.Player, message, type);
+                    break;
+                default:
+                    session.FieldManager.SendChat(session.Player, message);
+                    break;
```
AlanMorel, line 37: Since you're using a switch here, you might as well explicitly pass in the type, otherwise you might as well just make all of this a single line of code if there's going to be nothing different between them

MapleServer2/Packets/ChatPacket.cs
```diff
@@ -3,9 +3,12 @@
...
 using MapleServer2.Types;
 
-namespace MapleServer2.Packets {
-    public static class ChatPacket {
-        public static Packet Send(Player player, string message, ChatType type = ChatType.All) {
+namespace MapleServer2.Packets
+{
+    public static class ChatPacket
+    {
+        public static Packet Send(Player player, string message, ChatType type = ChatType.All)
```
AlanMorel, line 10: I don't think we should have a default ChatType, should be explicitly passed in

MapleServer2/Packets/ChatPacket.cs
```diff
@@ -16,16 +19,23 @@ public static class ChatPacket {
...
                     break;
                 case ChatType.Super:
-                    pWriter.WriteInt(); // item id?
+                    pWriter.WriteInt(20800017); // item id?
                     break;
                 case ChatType.UnknownPurple:
                     pWriter.WriteLong(); // char id?
                     break;
+                case ChatType.World:
+                    pWriter.WriteInt(); // Merets
```
AlanMorel, line 34: Was this tested? How do you know this is merets?

AlanMorel (changes requested):

MapleServer2/Packets/ChatPacket.cs
Sparkymod: True, is working for now, without sending anything, i was testing if it needs the merets in order to be send the message

AlanMorel:
Was this PR even tested at all? I noticed you added an extra parameter to `session.FieldManager.SendChat` but didn't include it in this PR, so merging this will break everybody's emulator

Sparkymod:
> Was this PR even tested at all? I noticed you added an extra parameter to `session.FieldManager.SendChat` but didn't include it in this PR, so merging this will break everybody's emulator

Yes, it was tested and works

Sparkymod:
Yeah, i was missing a file

> Was this PR even tested at all? I noticed you added an extra parameter to `session.FieldManager.SendChat` but didn't include it in this PR, so merging this will break everybody's emulator

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
rhydev: Also don't use `var`

rhydev (commented):
Looks good, just a couple style issues.

PR #12, opened 2020-12-16
Feat: mail

rhydev:
-Added Mail type to hold mail data
-Added Mailbox type to hold and manage mail
-Added handlers and packet files for mail
-Handles notify, send, open, read, collect, batch read, batch collect
-Blackmarket mail is not handled yet
-Added Item packet, not sure if this is only used for mail or not
-Added tick count to AccountStorage to fix timestamp issues
-Added new tool GuidGenerator to handle creating new uids
-Added a Mailbox to the GameSession
-Added test data to FieldEnterHandler with different types of mail

### DIFF
--- a/MapleServer2/PacketHandlers/Common/ResponseVersionHandler.cs
+++ b/MapleServer2/PacketHandlers/Common/ResponseVersionHandler.cs
@@ -5,6 +5,7 @@ using MapleServer2.Network;
 using MapleServer2.Packets;
 using MapleServer2.Servers.Game;
 using MapleServer2.Servers.Login;
+using MapleServer2.Data;
 using Microsoft.Extensions.Logging;
 
 namespace MapleServer2.PacketHandlers.Common {
@@ -14,6 +15,9 @@ namespace MapleServer2.PacketHandlers.Common {
         public ResponseVersionHandler(ILogger<ResponseVersionHandler> logger) : base(logger) { }
 
         public override void Handle(LoginSession session, PacketReader packet) {
+            // Sync the account state TickCount
+            AccountStorage.TickCount = Environment.TickCount;
+
             HandleCommon(session, packet);
 
             session.Send(RequestPacket.Login());
--- a/MapleServer2/PacketHandlers/Game/MailHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/MailHandler.cs
@@ -1,8 +1,12 @@
+using System;
+using System.Collections.Generic;
 using Microsoft.Extensions.Logging;
 using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
 using MapleServer2.Servers.Game;
 using MapleServer2.Packets;
+using MapleServer2.Types;
+using MapleServer2.Tools;
 
 namespace MapleServer2.PacketHandlers.Game
 {
@@ -41,7 +45,11 @@ namespace MapleServer2.PacketHandlers.Game
 
         private void HandleOpen(GameSession session, PacketReader packet)
         {
-            MailPacket.Open(session);
+            session.Mailbox.ClearExpired();
+
+            session.Send(MailPacket.StartOpen());
+            session.Send(MailPacket.Open(session.Mailbox.Box));
+            session.Send(MailPacket.EndOpen());
         }
 
         private void HandleSend(GameSession session, PacketReader packet)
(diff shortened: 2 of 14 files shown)


MapleServer2/PacketHandlers/Common/ResponseVersionHandler.cs
```diff
@@ -14,6 +15,9 @@ public class ResponseVersionHandler : CommonPacketHandler {
         public ResponseVersionHandler(ILogger<ResponseVersionHandler> logger) : base(logger) { }
 
         public override void Handle(LoginSession session, PacketReader packet) {
+            // Sync the account state TickCount
+            AccountStorage.TickCount = Environment.TickCount;
```
Sparkymod, line 19: Theres a ServerTick at GameSession.cs that does this, in any case are the same, you can use session.ServerTick , or there's a reason to make it in another class?

MapleServer2/Types/Player.cs
```diff
@@ -198,8 +199,8 @@ public static Player NewCharacter(byte gender, /*Job jobType*/ int job, string n
             {
                 SkillTabs = skillTabs,
                 AccountId = 0x1111111111111111,
-                CharacterId = BitConverter.ToInt64(Guid.NewGuid().ToByteArray(), 0),
-                CreationTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds(),
+                CharacterId = GuidGenerator.Long(),
+                CreationTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds() + AccountStorage.TickCount,
```
Sparkymod, line 203: the same as above in the same class, could use session.ServerTick ?

MapleServer2/PacketHandlers/Game/MailHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+            int count = packet.ReadInt();
+
+            int id = 0;
+            for (int i = 0; i < count; i++)
+            {
+                id = packet.ReadInt();
+                packet.ReadInt();
+
+                session.Send(MailPacket.Read(session, id));
+                MailPacket.Collect(session, id);
```
AlanMorel, line 95: I am not sure a method inside `MailPacket` should be making inventory calls or sending multiple packets. Sounds like a new class might be needed to put this kind of mail management in

MapleServer2/Packets/Helpers/MailPacketHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            pWriter.WriteInt(0);
+            pWriter.WriteLong(0);
+            pWriter.WriteInt(0);
+            pWriter.WriteInt(0);
+            pWriter.WriteZero(21);
+            pWriter.WriteInt(0);
+            pWriter.WriteInt(0);
+            pWriter.WriteLong(mail.ReadTimestamp > 0 ? mail.ReadTimestamp + AccountStorage.TickCount : 0); // Read timestamp
+            pWriter.WriteLong(mail.SentTimestamp + 864000000); // Time left
```
AlanMorel, line 65: Perhaps a comment should be made about what that number adds up to be, or break up the number into recognizable time chunks (like `60 * 60 * 24`, etc) since right now it feels magic number-y, same with line 91

MapleServer2/Packets/MailPacket.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            pWriter.WriteByte(14); // Mode for mail notification
+            pWriter.WriteInt(session.Mailbox.GetUnread()); // Count of unread mail
+            pWriter.WriteByte(); // Unknown
+            pWriter.WriteInt(); // Unknown maybe repeat of count?
+
+            return pWriter;
+        }
+
+        public static void Open(GameSession session)
```
AlanMorel, line 27: Same as before, I am not sure we should be having methods inside a packet class sending multiple packets

MapleServer2/Types/Mailbox.cs
```diff
@@ -0,0 +1,59 @@
...
+        public List<Item> Collect(int id)
+        {
+            int index = Box.FindIndex(x => x.Uid == id);
+            List<Item> items = Box[index].Items;
+            Box[index].Collect(null);
+
+            return items;
+        }
+
+        public int GetUnread()
```
AlanMorel, line 45: I feel like the method names here should better describe what the method is actually doing, for example `Read` might be better as `GetReadTimestamp`, `Collect` as `GetMailById` and `GetUnread` as `GetUnreadCount` since that can easily be interpreted as getting all the unread mail, and not simply the count of them.

AlanMorel (changes requested):

MapleServer2/Types/Mailbox.cs
rhydev: `Read` is setting the `ReadTimestamp` though, and `Collect` sets the `Items`. `GetUnread` could definitely be renamed to `GetUnreadCount` though.

MapleServer2/Types/Player.cs
rhydev: `Player` does not have access to `GameSession` because it is a Type so for the 2 hardcoded characters it cannot use a `GameSession` property, furthermore we want the `TickCount` from when the client first connected, the `session.ServerTick` updates over time.

MapleServer2/PacketHandlers/Common/ResponseVersionHandler.cs
rhydev: It could be moved as a new property in `GameSession`, and then when `NewCharacter` is called it would need to have the `CreationTime` passed as a property.

MapleServer2/PacketHandlers/Game/MailHandler.cs
rhydev: I agree that it's a bit strange to send multiple packets in a packet method, perhaps the multiple packets should be handled in the handler?

MapleServer2/Packets/Helpers/MailPacketHelper.cs
rhydev: This is just 86400 * 10000 where 86400 is the number of seconds in a day, perhaps it would be better to just have it written in a comment? `// Time left + 10000 days`?

MapleServer2/PacketHandlers/Common/ResponseVersionHandler.cs
rhydev: After experimentation, unfortunately it cannot be moved to `GameSession`, it would need to be pulled from `LoginSession` which breaks current implementation of generics for handling packets.

MapleServer2/PacketHandlers/Game/MailHandler.cs
AlanMorel: Yeah I think that would work just fine

MapleServer2/Packets/Helpers/MailPacketHelper.cs
AlanMorel: I think a 10,000 days comment would be enough

PR #16, opened 2020-12-17
Added chat functionality (whispers, world chat) and BroadcastAll

integralcoder95:
Feat: Properly broadcasting chat
Feat: Added BroadcastAll and BroadcastPacketAll to MapleServer
This lets you access all players currently online on the server (not just the current field)

### DIFF
--- a/MapleServer2/MapleServer.cs
+++ b/MapleServer2/MapleServer.cs
@@ -11,6 +11,8 @@ using NLog;
 
 namespace MapleServer2 {
     public static class MapleServer {
+
+        private static GameServer gameServer;
         public static void Main(string[] args) {
             // No DI here because MapleServer is static
             Logger logger = LogManager.GetCurrentClassLogger();
@@ -23,7 +25,7 @@ namespace MapleServer2 {
 
             IContainer gameContainer = GameContainerConfig.Configure();
             using ILifetimeScope gameScope = gameContainer.BeginLifetimeScope();
-            var gameServer = gameScope.Resolve<GameServer>();
+            gameServer = gameScope.Resolve<GameServer>();
             gameServer.Start();
 
             // Input commands to the server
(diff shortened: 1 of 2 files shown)


MapleServer2/MapleServer.cs
```diff
@@ -58,6 +60,25 @@ public static class MapleServer {
             }
         }
 
+        public static void BroadcastPacketAll(Packet packet, GameSession sender = null)
```
Sparkymod, line 63: Instead of using it here, calling methods on the same GameSession would be good for now, or make a new Class to handle Request of Sessions.

MapleServer2/MapleServer.cs
```diff
@@ -11,6 +11,8 @@
 
 namespace MapleServer2 {
     public static class MapleServer {
+
+        private static GameServer gameServer;
```
Sparkymod, line 15: If you want to use him, instead declare him on gamesession as `private readonly static GameServer gameServer;`

MapleServer2/MapleServer.cs
Sparkymod: Why don't use `public void` instead of `public static void` ? in this way you can add the method in the GameSession and can be called on the UserChatHandler.

Sparkymod: Checking again the code, FieldManager has the same methods, instead of making new one, should be use the one that FieldManager has

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -21,7 +22,44 @@ public override void Handle(GameSession session, PacketReader packet)
...
+                    break;
+                case ChatType.GuildNotice:
+                case ChatType.Guild:
+                    //TODO: Send to all in guild
+                    break;
+                case ChatType.Party:
+                    //TODO: Send to all in party
+                    break;
+                case ChatType.WhisperTo:
+                    Types.Player recipientPlayer = null;
```
AlanMorel, line 42: While not wrong, can you import `Types` so you can just use `Player` here?

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -21,7 +22,44 @@ public override void Handle(GameSession session, PacketReader packet)
...
+                    MapleServer.BroadcastAll(pSession => {
+                        if (pSession.Player.Name == recipient)
+                        {
+                            pSession.Send(ChatPacket.Send(session.Player, message, ChatType.WhisperFrom));
```
AlanMorel, line 46: The player who sent the whisper should only get the packet if the recipient was found, so maybe move this down to line 52 and send both packets together?

AlanMorel (changes requested):

MapleServer2/MapleServer.cs
integralcoder95: The Broadcast methods in FieldManager are only for the current field, it does not include all players in the server. I'll look into making a Session controller, but not sure how to implement keeping track of all sessions outside of GameServer

Sparkymod:
This should work for now, but improvement should be made in the future about hanlding request for game session and login session

PR #18, opened 2020-12-18
Simplifying account/characters, improving server_enter packet

AlanMorel:


MapleServer2/Network/PacketRouter.cs
```diff
@@ -31,7 +30,6 @@ public class PacketRouter<T> where T : Session {
 
         private void Register(ImmutableDictionary<RecvOp, IPacketHandler<T>>.Builder builder,
                 IPacketHandler<T> packetHandler) {
-            logger.Debug($"Registered {packetHandler}");
```
Sparkymod, line 34: Why removing this?

Sparkymod (commented):
All seems good but just pointing out the console notice of packets registered

MapleServer2/Network/PacketRouter.cs
AlanMorel: Just creates noise in the console on every start up for me, doesn't seem like information you need to know every time the server starts, the handler will register if you follow the structure of all the other handlers

PR #19, opened 2020-12-18
map command

Andoriii:


MapleServer2/Data/Static/MapEntityStorage.cs
```diff
@@ -24,5 +26,21 @@ public static class MapEntityStorage {
...
+        public static bool HasPortals(int mapId)
+        {
+            List<MapPortal> items = portals.GetValueOrDefault(mapId);
+            return items?.Count > 0;
+        }
+
+        public static MapPortal GetFirstPortal(int mapId)
+        {
+            List<MapPortal> items = portals.GetValueOrDefault(mapId);
+            if (items.Count > 0)
```
AlanMorel, line 39: Maybe do the same null check you're doing here as on line 33?

AlanMorel (changes requested):

PR #22, opened 2020-12-19
Create player storage, tie player with their own session

AlanMorel:


Sparkymod (approved):
Looks good, nothing to add

PR #24, opened 2020-12-19
Allow session-specific actions upon ending, end LoginSession when logging to game

AlanMorel:


MapleServer2/Servers/Login/LoginSession.cs
```diff
@@ -17,5 +17,9 @@ public class LoginSession : Session {
         public int GetToken() {
             return rng.Next();
         }
+
+        public override void EndSession()
+        {
```
Sparkymod, line 22: what is this doing ?

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -129,7 +129,6 @@ public void RemovePlayer(GameSession sender, IFieldObject<Player> player)
             {
                 session.Send(FieldPacket.RemovePlayer(player));
             });
-            GameServer.Storage.RemovePlayer(player.Value);
```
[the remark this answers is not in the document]
AlanMorel: This code was removing players from the PlayerStorage if they left the map, which is incorrect behavior

MapleServer2/Servers/Login/LoginSession.cs
AlanMorel: I added an abstract method in `Session` so any class that implements it needs to have an override, even if we don't need it to do anything for now

Sparkymod (approved):

Sparkymod (approved):

PR #25, opened 2020-12-19
Fix: skill tree

rhydev:
-Added new attribute "split" that stores the amount of skills left before sending this number


Sparkymod (approved):
Looks good

PR #40, opened 2020-12-22
Removed metadata from gitignore

Andoriii:


.gitignore
```diff
@@ -14,3 +14,4 @@ Maple2Storage/obj/
 MaplePacketLib2/obj/
 MapleServer2/obj/
 MapleServer2/Properties/launchSettings.json
+/Maple2Storage/Resources
```
AlanMorel, line 17: Can you remove the slash at the beginning?

AlanMorel (changes requested):

.gitignore
Andoriii: Done

PR #43, opened 2020-12-24
Adding: Skill to GameDataParser, Fixing: nullable values & other fix

Sparkymod:
- Adding *./File/Export* for set MetadataExports.
- Moving redundant method to separate files.
- Create a global Class to set the path of the files to read.
- Fixing nullable objects on ItemParser and MapEntityParser.
- Adding SkillParser, SkillMetadata & SkillMetadataStorage

### DIFF
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -12,9 +12,9 @@ namespace GameDataParser.Files.Export
 
         public void Export()
         {
-            string headerFile = SetPath.XML_PATH.Replace(".m2d", ".m2h");
+            string headerFile = VariableDefines.XML_PATH.Replace(".m2d", ".m2h");
             List<PackFileEntry> files = FileList.ReadFile(File.OpenRead(headerFile));
-            MemoryMappedFile memFile = MemoryMappedFile.CreateFromFile(SetPath.XML_PATH);
+            MemoryMappedFile memFile = MemoryMappedFile.CreateFromFile(VariableDefines.XML_PATH);
 
             // Parse and save some item data from xml file
             List<ItemMetadata> items = ItemParser.Parse(memFile, files);
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -7,7 +7,6 @@ using Maple2Storage.Types;
 
 namespace GameDataParser.Files.Export
 {
-
     public class ItemMetadataExport
     {
 
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -9,11 +9,13 @@ namespace GameDataParser.Files.Export
 {
     public class ItemMetadataExport
     {
+        private readonly SetPath setPath= new SetPath();
+
         public void Export()
         {
-            string headerFile = SetPath.XML_PATH.Replace(".m2d", ".m2h");
+            string headerFile = setPath.XML_PATH.Replace(".m2d", ".m2h");
             List<PackFileEntry> files = FileList.ReadFile(File.OpenRead(headerFile));
-            var memFile = MemoryMappedFile.CreateFromFile(SetPath.XML_PATH);
+            MemoryMappedFile memFile = MemoryMappedFile.CreateFromFile(setPath.XML_PATH);
 
             // Parse and save some item data from xml file
             List<ItemMetadata> items = ItemParser.Parse(memFile, files);
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -7,6 +7,7 @@ using Maple2Storage.Types;
 
 namespace GameDataParser.Files.Export
 {
+
     public class ItemMetadataExport
     {
 
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -1,0 +1,23 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using GameDataParser.Crypto.Common;
+using GameDataParser.Parsers;
+using Maple2Storage.Types;
+
+namespace GameDataParser.Files.Export
+{
+    public class ItemMetadataExport
+    {
+        public void Export()
+        {
+            string headerFile = SetPath.XML_PATH.Replace(".m2d", ".m2h");
+            List<PackFileEntry> files = FileList.ReadFile(File.OpenRead(headerFile));
+            var memFile = MemoryMappedFile.CreateFromFile(SetPath.XML_PATH);
+
+            // Parse and save some item data from xml file
+            List<ItemMetadata> items = ItemParser.Parse(memFile, files);
+            ItemParser.Write(items);
+        }
+    }
+}
--- a/GameDataParser/Files/Export/ItemMetadataExport.cs
+++ b/GameDataParser/Files/Export/ItemMetadataExport.cs
@@ -9,13 +9,12 @@ namespace GameDataParser.Files.Export
 {
     public class ItemMetadataExport
     {
-        private readonly SetPath setPath= new SetPath();
 
         public void Export()
         {
-            string headerFile = setPath.XML_PATH.Replace(".m2d", ".m2h");
+            string headerFile = SetPath.XML_PATH.Replace(".m2d", ".m2h");
             List<PackFileEntry> files = FileList.ReadFile(File.OpenRead(headerFile));
-            MemoryMappedFile memFile = MemoryMappedFile.CreateFromFile(setPath.XML_PATH);
+            MemoryMappedFile memFile = MemoryMappedFile.CreateFromFile(SetPath.XML_PATH);
 
             // Parse and save some item data from xml file
             List<ItemMetadata> items = ItemParser.Parse(memFile, files);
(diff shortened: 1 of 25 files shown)


GameDataParser/Files/SetPath.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace GameDataParser.Files
+{
+    public class SetPath
+    {
+        public const string XML_PATH = @"H:\MS2 Server\GMS2_final_version\maplestory2\appdata\Data\Xml.m2d";
```
AlanMorel, line 5: is this going to work for other people?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -0,0 +1,130 @@
...
+                    motions.Add(new SkillMotion(sequenceName, motionEffect, strTagEffects));
+                    metadata.SkillLevel.Add(new SkillLevel(level, spirit, upgradeLevel, upgradeSkillId, upgradeSkillLevel, motions));
+                }
+
+                skillList.Add(metadata);
+            }
+            return skillList;
+        }
+
+
```
AlanMorel, line 115: Double empty line?

Maple2Storage/Types/SkillMetadata.cs
```diff
@@ -0,0 +1,195 @@
...
+        public readonly List<SkillLevel> SkillLevel;
+
+        public SkillMetadata()
+        {
+            this.SkillLevel = new List<SkillLevel>();
+        }
+
+        public SkillMetadata(int skillId)
+        {
+
```
AlanMorel, line 37: why this blank line here?

Maple2Storage/Types/SkillMetadata.cs
```diff
@@ -0,0 +1,195 @@
...
+        }
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(SkillId, Feature, AttackType, Type, SubType, RangeType, Element, SkillLevel);
+        }
+        public static bool operator ==(SkillMetadata left, SkillMetadata right)
+        {
+            return Equals(left, right);
+        }
+        public static bool operator !=(SkillMetadata left, SkillMetadata right)
```
AlanMorel, line 70: Can we please leave a blank line in between methods?

Maple2Storage/Types/SkillMetadata.cs
```diff
@@ -0,0 +1,195 @@
...
+        public static bool operator ==(SkillLevel left, SkillLevel right)
+        {
+            return Equals(left, right);
+        }
+
+        public static bool operator !=(SkillLevel left, SkillLevel right)
+        {
+            return !Equals(left, right);
+        }
+        public override string ToString() =>
```
AlanMorel, line 142: Leave a blank line in between methods?

Maple2Storage/Types/SkillMetadata.cs
```diff
@@ -0,0 +1,195 @@
...
+        public static bool operator ==(SkillMotion left, SkillMotion right)
+        {
+            return Equals(left, right);
+        }
+
+        public static bool operator !=(SkillMotion left, SkillMotion right)
+        {
+            return !Equals(left, right);
+        }
+        public override string ToString() =>
```
AlanMorel, line 192: Once again, space in between methods

AlanMorel (changes requested):

GameDataParser/Files/SetPath.cs
Sparkymod: For other people to work they have to set this manually to their actual path of the game, i can add a comment to let people know what file needs on those path

eotetb: We really want to avoid putting in hardcoded absolute paths within the source code. If anything, at least include a relative path that people can put into their repo folder and add that folder to the git exceptions list.

GameDataParser/Files/Export/ItemMetadataExport.cs
```diff
@@ -9,11 +9,13 @@ namespace GameDataParser.Files.Export
 {
     public class ItemMetadataExport
     {
+        private readonly SetPath setPath= new SetPath();
```
AlanMorel, line 12: Space before the `=`?

AlanMorel: Also, if this is only used in the `Export()` method, why is it defined outside of it?

GameDataParser/Files/SetPath.cs
```diff
@@ -1,10 +1,11 @@
-﻿namespace GameDataParser.Files
+﻿using System.Configuration;
+
+namespace GameDataParser.Files
 {
     public class SetPath
```
AlanMorel, line 5: Why wouldn't this all be static?

AlanMorel (changes requested):

GameDataParser/Files/Export/ItemMetadataExport.cs
Sparkymod: You're right this could be ask without calling it outside and set the class to static

GameDataParser/Files/SetPath.cs
```diff
@@ -2,10 +2,10 @@
 
 namespace GameDataParser.Files
 {
-    public class SetPath
+    public static class SetPath
```
AlanMorel, line 5: Any reason this is called `SetPath`? It sounds like it would be a method name, not the name of a class

AlanMorel (changes requested):

GameDataParser/Files/SetPath.cs
Sparkymod: Didnt came with something else, but initially was treaten as method but i move it to a class, perhaps could be call, VariableSettings?

AlanMorel: Sure? I mean anything is better than `SetPath`, because the current name doesn't really make any sense at all

GameDataParser/Parsers/ItemParser.cs
```diff
@@ -3,42 +3,56 @@
...
+using System.Configuration;
 
-namespace GameDataParser.Parsers {
-    public static class ItemParser {
-        private const string OUTPUT = "ms2-item-metadata";
+namespace GameDataParser.Parsers
+{
+    public static class ItemParser
+    {
+        private static readonly string OUTPUT = ConfigurationManager.AppSettings["OUTPUT"] + "/ms2-item-metadata";
```
AlanMorel, line 17: Is there a reason we have a file for some of these settings but all, like in this case, `OUTPUT`?

AlanMorel (changes requested):

GameDataParser/Parsers/ItemParser.cs
Sparkymod: Becuase it will export the file into `/bin/debug/netappcore3.1/Resource` and instead, setting it to export on `GameDataParser/Resources`, setting at the config, will let people just to rename the last string to be the name of the file

GameDataParser/Parsers/ItemParser.cs
```diff
@@ -69,11 +68,11 @@ public static List<ItemMetadata> Parse(MemoryMappedFile m2dFile, IEnumerable<Pac
 
         public static void Write(List<ItemMetadata> items)
         {
-            using (FileStream writeStream = File.OpenWrite(OUTPUT))
+            using (FileStream writeStream = File.OpenWrite(VariableDefines.OUTPUT+ "ms2-item-metadata"))
```
AlanMorel, line 71: Space after `OUTPUT`?

AlanMorel (changes requested):

GameDataParser/Parsers/ItemParser.cs
```diff
@@ -53,30 +66,37 @@ public static class ItemParser {
...
-        public static void Write(List<ItemMetadata> items) {
-            using (FileStream writeStream = File.OpenWrite(OUTPUT)) {
+        public static void Write(List<ItemMetadata> items)
+        {
+            using (FileStream writeStream = File.OpenWrite(VariableDefines.OUTPUT + "ms2-item-metadata"))
+            {
                 Serializer.Serialize(writeStream, items);
             }
-            using (FileStream readStream = File.OpenRead(OUTPUT)) {
+            using (FileStream readStream = File.OpenRead(VariableDefines.OUTPUT + "ms2-item-metadata"))
```
AlanMorel, line 75: Is this `using` block doing anything?

GameDataParser/Parsers/MapEntityParser.cs
```diff
@@ -124,22 +146,26 @@ public static class MapEntityParser {
...
-        public static void Write(List<MapEntityMetadata> entities) {
-            using (FileStream writeStream = File.OpenWrite(OUTPUT)) {
+        public static void Write(List<MapEntityMetadata> entities)
+        {
+            using (FileStream writeStream = File.OpenWrite(VariableDefines.OUTPUT + "ms2-map-entity-metadata"))
+            {
                 Serializer.Serialize(writeStream, entities);
             }
-            using (FileStream readStream = File.OpenRead(OUTPUT)) {
+            using (FileStream readStream = File.OpenRead(VariableDefines.OUTPUT + "ms2-map-entity-metadata"))
```
AlanMorel, line 165: Same here

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -0,0 +1,128 @@
...
+            return skillList;
+        }
+
+        public static void Write(List<SkillMetadata> skills)
+        {
+            using (FileStream writeStream = File.OpenWrite(VariableDefines.OUTPUT + "ms2-skill-metadata"))
+            {
+                Serializer.Serialize(writeStream, skills);
+            }
+            using (FileStream readStream = File.OpenRead(VariableDefines.OUTPUT + "ms2-skill-metadata"))
```
AlanMorel, line 120: Same

AlanMorel (changes requested):

GameDataParser/Parsers/ItemParser.cs
Sparkymod: Serializing the file to export, and that is just use there, instead of calling at beging

AlanMorel: It looks like it's just reading the file but doing nothing with it

Sparkymod: To serialize an object it needs a buffer and the data, in this case is using the file that will be export for buffer and then pasing the data.

PR #54, opened 2020-12-26
Fixing numerous warnings

AlanMorel:


Sparkymod (approved):
Seems good, just one note, all the `Write<T>(T)` the program will know what to type of class to write without the `<T>` using `Write(T)` ?

PR #94, opened 2021-01-02
Club Invites, Leaving, Renaming, and Buff settings

Zintixx:
I commented on what needs to be done still but the biggest two things that needs to be addressed is removing club members out of party members and disbanding/reassigning leadership. I wanted to at least get this out for any comments and feedback

### DIFF
--- a/MapleServer2/PacketHandlers/Game/ClubHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/ClubHandler.cs
@@ -19,8 +19,11 @@ namespace MapleServer2.PacketHandlers.Game
         {
             Create = 0x1,
             Join = 0x3,
-            Invite = 0x6,
+            SendInvite = 0x6,
             InviteResponse = 0x8,
+            Leave = 0xA,
+            Buff = 0xD,
+            Rename = 0xE,
         }
 
         public override void Handle(GameSession session, PacketReader packet)
@@ -35,11 +38,20 @@ namespace MapleServer2.PacketHandlers.Game
                 case ClubMode.Join:
                     HandleJoin(session, packet);
                     break;
-                case ClubMode.Invite:
-                    HandleInvite(session, packet);
+                case ClubMode.SendInvite:
+                    HandleSendInvite(session, packet);
                     break;
                 case ClubMode.InviteResponse:
                     HandleInviteResponse(session, packet);
+                    break;
+                case ClubMode.Leave:
+                    HandleLeave(session, packet);
+                    break;
+                case ClubMode.Buff:
+                    HandleBuff(session, packet);
+                    break;
+                case ClubMode.Rename:
+                    HandleRename(session, packet);
                     break;
                 default:
                     IPacketHandler<GameSession>.LogUnknownMode(mode);
@@ -49,6 +61,7 @@ namespace MapleServer2.PacketHandlers.Game
 
         private void HandleCreate(GameSession session, PacketReader packet)
         {
+            // TODO fix creating for a party of more than 2. Currently if a member does not respond, despite atleast one other member accepting, it does not get created.
             Party party = GameServer.PartyManager.GetPartyByLeader(session.Player);
             if (party == null)
             {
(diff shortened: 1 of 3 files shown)


MapleServer2/PacketHandlers/Game/ClubHandler.cs
```diff
@@ -84,7 +96,7 @@ private void HandleJoin(GameSession session, PacketReader packet)
             {
                 foreach (Player member in party.Members)
                 {
-                    party.BroadcastPacketParty(ClubPacket.UpdatePlayerClubList(member, club.Id));
+                    party.BroadcastPacketParty(ClubPacket.UpdatePlayerClubList(session.Player, club));
```
AlanMorel, line 99: This seems wrong, this is making the packet send multiple times but only with the same person

MapleServer2/PacketHandlers/Game/ClubHandler.cs
```diff
@@ -104,22 +116,123 @@ private void HandleJoin(GameSession session, PacketReader packet)
...
+                return;
+            }
+
+            Club club = GameServer.ClubManager.GetClubById(clubId);
+            if (club == null)
+            {
+                return;
+            }
+
+            session.Send(ClubPacket.InviteSentReceipt(clubId, other.Session.Player));
```
AlanMorel, line 136: I think `other.Session.Player` can be simplified to just `other`

MapleServer2/PacketHandlers/Game/ClubHandler.cs
```diff
@@ -104,22 +116,123 @@ private void HandleJoin(GameSession session, PacketReader packet)
...
+            if (other == null)
+            {
+                return;
+            }
+
+            if (response == 1)
+            {
+                club.Leader.Session.Send(ClubPacket.LeaderInviteResponse(clubId, invitee, response));
+                club.BroadcastPacketClub(ClubPacket.ConfirmInvite(clubId, other.Session.Player, response, clubLeader));
+                other.Session.Send(ClubPacket.InviteResponse(clubId, clubName, clubLeader, session.Player, response));
```
AlanMorel, line 164: This should be updated to just take `club` instead of `clubId, clubName, clubLeader`, all of these can just be taken from the `club` itself

MapleServer2/Packets/ClubPacket.cs
```diff
@@ -50,11 +97,11 @@ public static Packet CreateClub(Party party, Club club)
             pWriter.WriteInt();
             pWriter.WriteInt();
             pWriter.WriteLong();
-            pWriter.WriteByte((byte)party.Members.Count);
+            pWriter.WriteByte((byte)party.Members.Count); //loop
 
             foreach (Player member in party.Members)
             {
-                pWriter.WriteByte((byte)party.Members.Count);
+                pWriter.WriteByte(0x2);
```
AlanMorel, line 104: So this is always 0x2 no matter what?

MapleServer2/Packets/ClubPacket.cs
```diff
@@ -73,88 +120,216 @@ public static Packet CreateClub(Party party, Club club)
...
+            return pWriter;
+        }
+
+        public static Packet Invite(Club club, Player other)
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.CLUB);
+            pWriter.WriteByte((byte)ClubPacketMode.Invite);
+            pWriter.WriteLong(club.Id);
+            pWriter.WriteUnicodeString(club.Name);
+            pWriter.WriteUnicodeString(club.Leader.Name); //clubLeader
```
AlanMorel, line 145: Comment here can be removed

AlanMorel (changes requested):

MapleServer2/Packets/ClubPacket.cs
Zintixx: Yeah, I was testing this and it appears that 0x2 is indicating the start of a loop, not the amount of times it has to loop/which loop it's at.

newgithubaccountlol (approved):
Looks good. Add spaces after variable casts, but otherwise good.

PR #100, opened 2021-01-04
Guild creation

Zintixx:
Start of guild system. A lot of names for the handles/packets are placeholders until it starts getting more fleshed out.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -18,9 +18,9 @@ namespace MapleServer2.PacketHandlers.Game
         {
             Create = 0x1,
             CheckIn = 0xF,
+            GuildNotice = 0x3E,
             GuildWindow = 0x54,
             List = 0x55,
-            GuildNotice = 0x3E,
             EnterHouse = 0x64,
             GuildDonate = 0x6E,
         }
@@ -61,7 +61,12 @@ namespace MapleServer2.PacketHandlers.Game
         {
             string guildName = packet.ReadUnicodeString();
 
-            if (session.Player.Mesos >= 2000)
+            if (session.Player.Mesos < 2000)
+            {
+                // TODO: Reject packets
+                return;
+            }
+            else
             {
                 session.Player.Mesos -= 2000;
 
@@ -74,11 +79,7 @@ namespace MapleServer2.PacketHandlers.Game
 
                 Guild newGuild = new(guildName, new List<Player> { session.Player });
                 GameServer.GuildManager.AddGuild(newGuild);
-            }
-            else
-            {
-                // TODO: Reject packets
-                return;
+
             }
 
         }
@@ -95,6 +96,15 @@ namespace MapleServer2.PacketHandlers.Game
             session.Send(GuildPacket.UpdatePlayerContribution(session.Player));
         }
 
+        private void HandleGuildNotice(GameSession session, PacketReader packet)
+        {
+            packet.ReadByte();
+            string notice = packet.ReadUnicodeString();
+
+            session.Send(GuildPacket.GuildNoticeConfirm(notice));
+            session.Send(GuildPacket.GuildNoticeChange(session.Player, notice)); // TODO: Change to Broadcast to Guild
+        }
+
         private void HandleGuildWindowRequest(GameSession session, PacketReader packet)
         {
             session.Send(GuildPacket.GuildWindow());
@@ -103,15 +113,6 @@ namespace MapleServer2.PacketHandlers.Game
         private void HandleList(GameSession session, PacketReader packet)
         {
             session.Send(GuildPacket.List());
-        }
-
-        private void HandleGuildNotice(GameSession session, PacketReader packet)
-        {
-            packet.ReadByte();
-            string notice = packet.ReadUnicodeString();
-
-            session.Send(GuildPacket.GuildNoticeConfirm(notice));
-            session.Send(GuildPacket.GuildNoticeChange(session.Player, notice)); // TODO: Change to Broadcast to Guild
         }
 
         private void HandleEnterHouse(GameSession session, PacketReader packet)
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -1,0 +1,125 @@
+﻿using System.Collections.Generic;
+using MaplePacketLib2.Tools;
+using MapleServer2.Constants;
+using MapleServer2.Packets;
+using MapleServer2.Servers.Game;
+using Microsoft.Extensions.Logging;
+using MapleServer2.Types;
+
+namespace MapleServer2.PacketHandlers.Game
+{
+    public class GuildHandler : GamePacketHandler
+    {
+        public override RecvOp OpCode => RecvOp.GUILD;
+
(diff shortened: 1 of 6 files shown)


MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        private enum GuildMode : byte
+        {
+            Create = 0x1,
+            CheckIn = 0xF,
+            EnterHouse = 0x64,
+            GuildWindow = 0x54,
+            List = 0x55,
+            GuildNotice = 0x3E,
+            GuildDonate = 0x6E,
```
AlanMorel, line 25: Can these be in order, in addition to the switch below, and the order of the methods in general

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+                    IPacketHandler<GameSession>.LogUnknownMode(mode);
+                    break;
+            }
+        }
+
+        private void HandleCreate(GameSession session, PacketReader packet)
+        {
+            string guildName = packet.ReadUnicodeString();
+
+            session.Player.Mesos -= 2000;
```
AlanMorel, line 64: A quick check if they have the mesos would be nice here, and you can just return if they don't have enough for now

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+            session.Send(GuildPacket.Invite(session.Player, guildName));
+            session.Send(GuildPacket.Create(guildName));
+            session.Send(GuildPacket.UpdateGuild(session, guildName));
+            session.Send(GuildPacket.Create2(session.Player, guildName));
+            session.Send(GuildPacket.Create3(session.Player, guildName));
+
+            Guild newGuild = new(guildName, new List<Player> { session.Player });
+            GameServer.GuildManager.AddGuild(newGuild);
+
+            newGuild.AddMember(session.Player);
```
AlanMorel, line 76: The guild should probably just add the member to the guild in the constructor since you always want this to happen

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+        private void HandleCheckIn(GameSession session, PacketReader packet)
+        {
+            session.Send(GuildPacket.CheckInConfirm());
+            // TODO: Send Guild Coins
+            // TODO: Can only check in once a day
+            session.Send(GuildPacket.UpdateGuildFunds());
+            session.Send(GuildPacket.UpdateGuildExp());
+            session.Send(GuildPacket.UpdateGuildFunds2());
+            session.Send(GuildPacket.UpdatePlayerContribution(session.Player));
+            session.Send(GuildPacket.UpdatePlayerContribution(session.Player));
```
AlanMorel, line 88: Why is this sent twice?

MapleServer2/Packets/GuildPacket.cs
```diff
@@ -0,0 +1,328 @@
...
+            Create3 = 0x20,
+            Unk1 = 0x24,
+            UpdateGuildExp = 0x31,
+            UpdateGuildFunds = 0x32,
+            UpdatePlayerContribution = 0x33,
+            GuildNoticeConfirm = 0x3E,
+            GuildWindow = 0x54,
+            List = 0x5A,
+            UpdateGuildFunds2 = 0x59, //unk
+            UpdatePlayerDonation = 0x6E,
```
AlanMorel, line 28: Can this be in order by value, same with the order of the methods below?

MapleServer2/Types/Guild.cs
```diff
@@ -0,0 +1,65 @@
...
+        public List<Player> Members { get; }
+
+        public Guild(string name, List<Player> gPlayers)
+        {
+            Id = GuidGenerator.Long();
+            Name = name;
+            MaxMembers = 60;
+            Leader = gPlayers.First();
+            Members = gPlayers;
+            Approval = false;
```
AlanMorel, line 27: This is where you can do `Mmembers.Add(Leader);`

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
Zintixx: First time it's sent is for guildExp contribution only. Second time is sent for guildFunds contributions only.

MapleServer2/Packets/GuildPacket.cs
```diff
@@ -0,0 +1,328 @@
...
+            pWriter.WriteInt(2);
+            pWriter.WriteInt(1);
+            pWriter.WriteLong(0);
+            pWriter.WriteInt(3);
+            pWriter.WriteInt(1);
+            pWriter.WriteLong(0);
+            pWriter.WriteInt(4);
+            pWriter.WriteInt(1);
+            pWriter.WriteLong(0);
+            pWriter.WriteInt(10001);
```
newgithubaccountlol, line 111: I'd make these into a loop of some sort also

MapleServer2/Packets/GuildPacket.cs
```diff
@@ -0,0 +1,328 @@
...
+            pWriter.WriteInt(0);
+            pWriter.WriteInt(0);
+            pWriter.WriteByte(0x1);
+            pWriter.WriteInt(0);
+            pWriter.WriteInt(0); // guild.Exp
+            pWriter.WriteInt(0); // guild.Funds
+            pWriter.WriteByte(0x0);
+            pWriter.WriteInt(0);
+            pWriter.WriteByte(0x0); // guild.Members.Count and also loop for each member ?
+            pWriter.WriteByte(0x6); // loop for guild ranks ?
```
newgithubaccountlol, line 79: I think this should be in a loop, maybe next change though. You can store the ranks in a list or an array.

MapleServer2/Types/Guild.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+using MaplePacketLib2.Tools;
+using MapleServer2.Packets;
+using MapleServer2.Servers.Game;
+using MapleServer2.Tools;
+
+namespace MapleServer2.Types
+{
+    public class Guild
+    {
```
newgithubaccountlol, line 12: do we want to add guild properties later (like guild map, level, rank names, etc?) Maybe drop a todo

newgithubaccountlol (commented):
Looks pretty good, I don't have any complaints other than lots of hardcoded packets but I understand this is a WIP

MapleServer2/Packets/GuildPacket.cs
Zintixx: these are part of the same guild skills loop

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -61,19 +61,26 @@ private void HandleCreate(GameSession session, PacketReader packet)
         {
             string guildName = packet.ReadUnicodeString();
 
-            session.Player.Mesos -= 2000;
+            if (session.Player.Mesos >= 2000)
```
AlanMorel, line 64: can you instead invert this check, then put the `TODO: reject packets` part here with the return? then the code after that doesn't need to live inside the `else` block

AlanMorel (changes requested):

PR #101, opened 2021-01-04
Feat: Interactable objects

rhydev:
- Added map object data to map Metadata
- Added new coord type for byte coords
- Added handler and packet classes for interactables

### DIFF
--- a/GameDataParser/Parsers/MapEntityParser.cs
+++ b/GameDataParser/Parsers/MapEntityParser.cs
@@ -19,22 +19,24 @@ namespace GameDataParser.Parsers
         {
             // Iterate over preset objects to later reference while iterating over exported maps
             Dictionary<string, string> mapObjects = new Dictionary<string, string>();
-
             foreach (PackFileEntry entry in entries)
             {
                 if (!entry.Name.StartsWith("flat/presets/presets object/"))
+                {
                     continue;
+                }
 
                 string objStr = entry.Name.ToLower();
 
                 if (string.IsNullOrEmpty(objStr))
+                {
                     continue;
+                }
                 if (mapObjects.ContainsKey(objStr))
                 {
                     Console.WriteLine($"Duplicate {entry.Name} was already added as {mapObjects[objStr]}");
                     continue;
                 }
-
 
                 XmlDocument document = m2dFile.GetDocument(entry.FileHeader);
 
@@ -104,8 +106,9 @@ namespace GameDataParser.Parsers
                         {
                             continue;
                         }
-                        CoordB coord = ParseCoordB(Regex.Match(nameCoord, @"[\-]?\d+[,]\s[\-]?\d+[,]\s[\-]?\d+").Value);
-                        metadata.Objects.Add(new MapObject(nameCoord, coord, int.Parse(mapObjects[modelName])));
+
+                        CoordB coord = ParseCoordB(coordMatch.Value);
+                        metadata.Objects.Add(new MapObject(coord, int.Parse(mapObjects[modelName])));
                     }
                 }
 
--- a/GameDataParser/Parsers/MapEntityParser.cs
+++ b/GameDataParser/Parsers/MapEntityParser.cs
@@ -26,8 +26,8 @@ namespace GameDataParser.Parsers
                     continue;
                 }
 
+                // Check if file is valid
                 string objStr = entry.Name.ToLower();
-
                 if (string.IsNullOrEmpty(objStr))
                 {
                     continue;
@@ -38,12 +38,12 @@ namespace GameDataParser.Parsers
                     continue;
                 }
 
+                // Parse XML
                 XmlDocument document = m2dFile.GetDocument(entry.FileHeader);
-
                 XmlElement root = document.DocumentElement;
-                string objectName = root.Attributes["name"].Value.ToLower();
-
                 XmlNodeList objProperties = document.SelectNodes("/model/property");
+
+                string objectName = root.Attributes["name"].Value.ToLower();
 
                 foreach (XmlNode node in objProperties)
                 {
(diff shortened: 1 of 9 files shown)


GameDataParser/Parsers/MapEntityParser.cs
```diff
@@ -17,6 +17,49 @@ public static class MapEntityParser
...
+                    continue;
+                }
+
+                string objStr = entry.Name.ToLower();
+
+                if (string.IsNullOrEmpty(objStr))
+                {
+                    continue;
+                }
+                if (mapObjects.ContainsKey(objStr))
```
AlanMorel, line 35: Can there be an empty line above this

GameDataParser/Parsers/MapEntityParser.cs
```diff
@@ -159,6 +218,16 @@ private static CoordS ParseCoord(string value)
             );
         }
 
+        private static CoordB ParseCoordB(string value)
```
AlanMorel, line 221: Can this method live inside `CoordB`? Seems like there's already methods there for creating new Coords from existing data

MapleServer2/PacketHandlers/Game/RequestCubeHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public RequestCubeHandler(ILogger<RequestCubeHandler> logger) : base(logger) { }
+
+        public override void Handle(GameSession session, PacketReader packet)
+        {
+            byte mode = packet.ReadByte();
+
+            switch (mode)
+            {
+                case 0x11: // Pickup item
```
AlanMorel, line 22: Can this be converted to use enums? You can look at `ClubHandler.cs` for a good example of this, then you can remove the comments here

MapleServer2/Packets/ResponseCubePacket.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    weaponId = mapObject.WeaponId;
+                    break;
+                }
+            }
+
+            pWriter.WriteShort(0x0011); // Mode
+            pWriter.WriteInt(session.FieldPlayer.ObjectId);
+            pWriter.Write(coords);
+            pWriter.WriteZero(1);
```
AlanMorel, line 36: Is this any different than just `pWriter.WriteByte()`?

MapleServer2/Packets/ResponseCubePacket.cs
```diff
@@ -0,0 +1,53 @@
...
+            foreach (MapObject mapObject in MapEntityStorage.GetObjects(session.Player.MapId))
+            {
+                if (mapObject.Coord.X == sCoords[0] && mapObject.Coord.Y == sCoords[1] && mapObject.Coord.Z == sCoords[2])
+                {
+                    weaponId = mapObject.WeaponId;
+                    break;
+                }
+            }
+
+            pWriter.WriteShort(0x0011); // Mode
```
AlanMorel, line 33: Can the modes be converted to enums? `ClubPacket.cs` has a good example of this

MapleServer2/Packets/ResponseCubePacket.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            // Convert to signed byte array
+            sbyte[] sCoords = Array.ConvertAll(coords, b => unchecked((sbyte) b));
+            // Default to rainbow tree
+            int weaponId = 18000004;
+
+            // Find matching mapObject
+            foreach (MapObject mapObject in MapEntityStorage.GetObjects(session.Player.MapId))
+            {
+                if (mapObject.Coord.X == sCoords[0] && mapObject.Coord.Y == sCoords[1] && mapObject.Coord.Z == sCoords[2])
```
AlanMorel, line 26: Don't need to change this in this PR but if this is used anywhere else in the future, it's probably better to make this a method that returns the boolean

AlanMorel (changes requested):

MapleServer2/Packets/ResponseCubePacket.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace MapleServer2.Packets
+{
+    public static class ResponseCubePacket
+    {
+        public static Packet Pickup(GameSession session, byte[] coords)
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.RESPONSE_CUBE);
+
+            // Convert to signed byte array
+            sbyte[] sCoords = Array.ConvertAll(coords, b => unchecked((sbyte) b));
```
AlanMorel, line 19: This should be done in the handler, and instead just pass `sCoords` in as a parameter

MapleServer2/Packets/ResponseCubePacket.cs
```diff
@@ -0,0 +1,53 @@
...
+            // Default to rainbow tree
+            int weaponId = 18000004;
```
AlanMorel, line 21: This line of code, and also the code below it finding the `mapObject` should also move to the handler, and have the packet method just get fed the final results

AlanMorel (changes requested):

MapleServer2/Packets/ResponseCubePacket.cs
rhydev: No, but it's more explicit, the byte here is always zero.

MapleServer2/Packets/ResponseCubePacket.cs
rhydev: The packet sends a `byte[]` not an `sbyte[]` so if I convert in handler I'd need to then unconvert in packet.

MapleServer2/Packets/ResponseCubePacket.cs
rhydev: Can't be moved because of `sbyte[]` issue as stated previously.

MapleServer2/Packets/ResponseCubePacket.cs
AlanMorel: You're only using `sCoords` here to set a new weapon id if it finds one, you can just send the packet method the finished item id, along with the `byte[]` that you need to write

GameDataParser/Parsers/MapEntityParser.cs
rhydev: I think it makes more sense without a newline since they are part of the same check block.

PR #59, opened 2020-12-29
Stat points increment and reset: implementation

nobodylikesaleech:
###### To be implemented: The remainder of the `WriteTotalStatPoints` method in `StatPointPacket` class and a functioning version of the `SendStatPacket` class. The incremented stats will currently not appear in the character window because the SendStat packet is not sent to update the character's information.
---
### Class: StatPointHandler 
Catches packets (RecvOp code 0x009A) requesting an increment in a stat value (ie +1 to Strength, Dex, Luck, Int, Health, or Crit Rate)
  * Packet format: `<mode byte> <stat index>`
    * Example packet: `02 00` (translated: increment the strength stat)
  * Mode byte - `02` represents the increment command
  * Mode byte - `03` represents the reset stat distribution command
  * Stat index - the indices represented by this byte include   
    * Strength - `00`
    * Dex - `01`
    * Int - `02`
    * Luck - `03`
    * Health - `04`
    * Crit Rate - `0x11` (17 in decimal form)
### Class: StatPointPacket 
Generates a packet (SendOp code 0x0046) in response to a request for StatPoints. Consists of the following methods:
  * WriteTotalAttributePoints
    * Method that sends the number of attribute points a character has gained and tracks of how each stat point was obtained (ie quest, trophy, exploration, prestige) 
    * This method is called when your character enters the field (ie logs in or new map)
      * ie. server emulator receives a `RESPONSE_FIELD_ENTER` (RecvOp) code from the client
      * result: server emulator sends back `STAT_POINT` packet (SendOp) with information to the client
      * client updates character window with total attribute points

  * WriteAttributePointDistribution
      * Method that sends the distribution of attribute points a character has
      * This packet is sent in response to an incrementation in the player's stats
      * After this stat distribution packet is sent, a follow up packet with op code 0x002F (SendStat) is sent. It contains the player's total stats and displays the updated information in the character's window.
### Class: StatDistribution 
Class that represents the Strength, Dex, Int, Luck, Health, and Crit Rate points a character has allocated to their stats.
* Allows Player to add points to their stats via `AddPoint(byte statType)` or reset their stat distribution via `ResetPoints()`

PR #103, opened 2021-01-05
Stat points increment and reset: implementation

nobodylikesaleech:
###### To be implemented: The remainder of the `WriteTotalStatPoints` method in `StatPointPacket` class and a functioning version of the `SendStatPacket` class. The incremented stats will currently not appear in the character window because the SendStat packet is not sent to update the character's information.
---
### Class: StatPointHandler 
Catches packets (RecvOp code 0x009A) requesting an increment in a stat value (ie +1 to Strength, Dex, Luck, Int, Health, or Crit Rate)
  * Packet format: `<mode byte> <stat index>`
    * Example packet: `02 00` (translated: increment the strength stat)
  * Mode byte - `02` represents the increment command
  * Mode byte - `03` represents the reset stat distribution command
  * Stat index - the indices represented by this byte include   
    * Strength - `00`
    * Dex - `01`
    * Int - `02`
    * Luck - `03`
    * Health - `04`
    * Crit Rate - `0x11` (17 in decimal form)
### Class: StatPointPacket 
Generates a packet (SendOp code 0x0046) in response to a request for StatPoints. 
Contains the following modes:
  * Mode byte - `00` requests the total stat points a player has
  * Mode byte - `01` requests the player's stat distribution

Consists of the following methods:
  * WriteTotalAttributePoints
    * Method that sends the number of attribute points a character has gained and tracks of how each stat point was obtained (ie quest, trophy, exploration, prestige) 
    * This method is called when your character enters the field (ie logs in or new map)
      * ie. server emulator receives a `RESPONSE_FIELD_ENTER` (RecvOp) code from the client
      * result: server emulator sends back `STAT_POINT` packet (SendOp) with information to the client
      * client updates character window with total attribute points
  * WriteAttributePointDistribution
      * Method that sends the distribution of attribute points a character has
      * This packet is sent in response to an incrementation in the player's stats
      * After this stat distribution packet is sent, a follow up packet with op code 0x002F (SendStat) is sent. It contains the player's total stats and displays the updated information in the character's window.
### Class: StatDistribution 
Class that represents the Strength, Dex, Int, Luck, Health, and Crit Rate points a character has allocated to their stats.
* Allows Player to add points to their stats via `AddPoint(byte statType)` or reset their stat distribution via `ResetPoints()`

### DIFF
--- a/MapleServer2/PacketHandlers/Game/StatPointHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/StatPointHandler.cs
@@ -1,0 +1,48 @@
+﻿using Microsoft.Extensions.Logging;
+using MaplePacketLib2.Tools;
+using MapleServer2.Constants;
+using MapleServer2.Servers.Game;
+using MapleServer2.Packets;
+using System;
+
+namespace MapleServer2.PacketHandlers.Game
+{
+    public class StatPointHandler : GamePacketHandler
+    {
+        public override RecvOp OpCode => RecvOp.STAT_POINT;
+
+        public StatPointHandler(ILogger<StatPointHandler> logger) : base(logger) { }
+
+        public override void Handle(GameSession session, PacketReader packet)
+        {
+            byte mode = packet.ReadByte();
+
+            switch (mode)
+            {
+                case 2: 
+                // Increment attribute point
+                    HandleStatIncrement(session, packet);
+                    // follow up SendStat packet to refresh updated stats
+                    break;
+                case 3:
+                // Reset attribute distribution
+                    HandleResetStatDistribtuion(session, packet);
+                    // follow up SendStat packet to refresh updated stats
+                    break;
+            }
+        }
+
+        private void HandleStatIncrement(GameSession session, PacketReader packet)
+        {
+            byte statTypeIndex = packet.ReadByte();
+            Console.WriteLine("statIndex value: " + statTypeIndex);
+            StatPointPacket.AddStatPoint(session.Player, statTypeIndex);
+            session.Send(StatPointPacket.WriteStatPointDistribution(session.Player));
+        }
+        private void HandleResetStatDistribtuion(GameSession session, PacketReader packet)
+        {
+            StatPointPacket.ResetStatPoints(session.Player);
+            session.Send(StatPointPacket.WriteStatPointDistribution(session.Player));
+        }
+    }
+}
--- a/MapleServer2/PacketHandlers/Game/StatPointHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/StatPointHandler.cs
@@ -22,12 +22,10 @@ namespace MapleServer2.PacketHandlers.Game
                 case 2: 
                 // Increment attribute point
                     HandleStatIncrement(session, packet);
-                    // follow up SendStat packet to refresh updated stats
                     break;
                 case 3:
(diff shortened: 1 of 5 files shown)


nobodylikesaleech:
Former PR: #59

MapleServer2/PacketHandlers/Game/StatPointHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+            StatPointMode mode = (StatPointMode) packet.ReadByte();
+
+            switch (mode)
+            {
+                case StatPointMode.Increment:
+                    HandleStatIncrement(session, packet);
+                    break;
+                case StatPointMode.Reset:
+                    HandleResetStatDistribution(session);
+                    break;
```
AlanMorel, line 33: Can you add 
```
default:
    IPacketHandler<GameSession>.LogUnknownMode(mode);
    break;
``` 
just like is seen inside `ClubHandler.cs`

MapleServer2/Packets/StatPointPacket.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            return pWriter;
+        }
+
+        public static void AddStatPoint(Player character, byte statType)
+        {
+            character.StatPointDistribution.AddPoint(statType);
+        }
+
+        public static void ResetStatPoints(Player character)
```
AlanMorel, line 56: These two lines of code (here and line 53) can just be called directly insider the handler since we only want packets methods inside of packet files

MapleServer2/Types/StatDistribution.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            this.AllocatedStats = AllocatedStats;
+        }
+
+        public void AddPoint(byte StatType)
+        {
+            int statCount;
+            if (AllocatedStats.TryGetValue(StatType, out statCount))
+            {
+                AllocatedStats[StatType] = statCount + 1;
```
AlanMorel, line 36: Would this work by just doing `AllocatedStats[StatType] += 1`? And if so, we can then remove the need for `statCount` altogether and just check for the existence of the stat

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/StatPointHandler.cs
nobodylikesaleech: Added!

MapleServer2/Packets/StatPointPacket.cs
nobodylikesaleech: This can be changed - originally, I wanted to go for readability, but removing these functions would be more concise and keep the code readable

nobodylikesaleech: Functions have been removed.

MapleServer2/Types/StatDistribution.cs
nobodylikesaleech: Changing my implementation as suggested - initially, I thought TryGetValue would be more efficient because it performs better when the key is frequently not found (as would be the case when a user starts with or resets to an empty stat distribution).

Since there are only 6 stats and players usually only allocate to 1 or 2 stats, ContainsKey would perform better and the implementation should be more readable to non C# coders.

PR #104, opened 2021-01-05
Minor fix: reverting working directory to SolutionDir

nobodylikesaleech:
Reverting the working directory to a generalized setting.


rhydev:
This was fixed in a prior commit, Properties should be ignored now.

PR #105, opened 2021-01-05
Feat: Server notices

rhydev:
- Created Notice packet class
- Added command to send server-wide notice
- Changed GameDataParsers to always create new file


Sparkymod (approved):
Looks good

PR #107, opened 2021-01-05
Feature Currency

angelotadeucci:
- Added ENUM CurrencyType
- Added Class Currency
- Added Class Wallet

Added two commands for testing purposes:

- setmesos / setmerets 

ISSUE: Missing packets to update the client for:

- ValorToken
- Treva
- Rue
- HaviFruit
- MesoToken

### DIFF
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -4,6 +4,8 @@
     {
         Meso,
         Meret,
+        GameMeret,
+        EventMeret,
         ValorToken,
         Treva,
         Rue,
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -1,4 +1,7 @@
-﻿namespace MapleServer2.Enums
+﻿using MapleServer2.Servers.Game;
+using MapleServer2.Types;
+
+namespace MapleServer2.Enums
 {
     public enum CurrencyType
     {
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -1,7 +1,4 @@
-﻿using MapleServer2.Servers.Game;
-using MapleServer2.Types;
-
-namespace MapleServer2.Enums
+﻿namespace MapleServer2.Enums
 {
     public enum CurrencyType
     {
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -6,7 +6,6 @@
         Meret,
         GameMeret,
         EventMeret,
-        RedMeret,
         ValorToken,
         Treva,
         Rue,
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -1,0 +1,31 @@
+﻿namespace MapleServer2.Enums
+{
+    public enum CurrencyType
+    {
+        Meso,
+        Meret,
+        ValorToken,
+        Treva,
+        Rue,
+        HaviFruit,
+        MesoToken,
+    }
+
+    /* All currencies from wiki:
+     
+    Merets
+    Mesos
+    Havi Fruit
+    Rue(Currency)
+    Treva
+    Valor Token
+
+    Blue Star(Currency)
+    Event Dungeon Coin
+    Guild Coins
+    Kay Coin
+    Maple Coin
+    Premium Coin
+    Red Star
+    */
+}
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -9,6 +9,13 @@
         Rue,
         HaviFruit,
         MesoToken,
+        BlueStar,
+        RedStar,
+        EventDungeonCoin,
+        GuildCoins,
+        KayCoin,
+        MapleCoin,
+        PremiumCoin,
     }
 
     /* All currencies from wiki:
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -17,22 +17,4 @@
         MapleCoin,
         PremiumCoin,
     }
-
-    /* All currencies from wiki:
-     
-    Merets
-    Mesos
-    Havi Fruit
-    Rue(Currency)
-    Treva
-    Valor Token
-
-    Blue Star(Currency)
-    Event Dungeon Coin
-    Guild Coins
-    Kay Coin
-    Maple Coin
-    Premium Coin
-    Red Star
-    */
 }
--- a/MapleServer2/Enums/CurrencyType.cs
+++ b/MapleServer2/Enums/CurrencyType.cs
@@ -6,17 +6,11 @@
         Meret,
         GameMeret,
         EventMeret,
+        RedMeret,
         ValorToken,
         Treva,
         Rue,
         HaviFruit,
         MesoToken,
-        BlueStar,
-        RedStar,
-        EventDungeonCoin,
-        GuildCoins,
-        KayCoin,
-        MapleCoin,
-        PremiumCoin,
     }
 }
--- a/MapleServer2/Tools/GameCommandActions.cs
+++ b/MapleServer2/Tools/GameCommandActions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Numerics;
 using Maple2Storage.Types;
 using Maple2Storage.Types.Metadata;
 using MapleServer2.Data.Static;
@@ -179,7 +180,6 @@ namespace MapleServer2.Tools
 
         private static long ParseLong(GameSession session, string s)
         {
-            long temp = -1;
             try
             {
                 return long.Parse(s);
@@ -187,16 +187,16 @@ namespace MapleServer2.Tools
             catch (FormatException)
             {
                 session.SendNotice("The input is not type long.");
-                return temp;
+                return -1;
             }
             catch (OverflowException)
             {
                 session.SendNotice("You entered a number too big or too small.");
-                return temp;
+                return -1;
             }
             catch (Exception)
             {
-                return temp;
+                return -1;
             }
         }
     }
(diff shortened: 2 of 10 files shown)


AlanMorel (changes requested):

MapleServer2/Tools/GameCommandActions.cs
```diff
@@ -170,5 +176,28 @@ private static bool TryParseCoord(string s, out CoordF result)
...
+            catch (OverflowException)
+            {
+                session.SendNotice("You entered a number too big or too small.");
+                return -1;
+            }
+            catch (Exception)
+            {
+                return -1;
+            }
+            return long.Parse(s);
```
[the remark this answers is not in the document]
rhydev: Can just return in the `try`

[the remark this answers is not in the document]
angelotadeucci:
```
private static long ParseLong(GameSession session, string s)
        {
            long temp = -1;
            try
            {
                return long.Parse(s);
            }
            catch (FormatException)
            {
                session.SendNotice("The input is not type long.");
                return temp;
            }
            catch (OverflowException)
            {
                session.SendNotice("You entered a number too big or too small.");
                return temp;
            }
            catch (Exception)
            {
                return temp;
            }
        }
```
This is good?

MapleServer2/Tools/GameCommandActions.cs
```diff
@@ -170,5 +176,28 @@ private static bool TryParseCoord(string s, out CoordF result)
             result = default;
             return false;
         }
+
+        private static long ParseLong(GameSession session, string s)
+        {
+            long temp = -1;
```
AlanMorel, line 182: Now you don't need this temp since you returned on line 185, you can just return `-1` directly

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public long Meso { get; set; }
+        public long Meret { get; set; }
+        public long ValorToken { get; set; }
+        public long Treva { get; set; }
+        public long Rue { get; set; }
+        public long HaviFruit { get; set; }
+        public long MesoToken { get; set; }
+
+        public Currency()
```
AlanMorel, line 16: I think the idea for this class was to have each individual currency represented by this single class, rather than one class representing them all. This class would also handle manipulating the amount, minus the packet sending, which can still be done by the `Player.cs` class

AlanMorel (changes requested):

MapleServer2/Types/Player.cs
```diff
@@ -91,16 +91,28 @@ public class Player
...
             StatDistribution StatPointDistribution = new StatDistribution();
-            Currency currency = new Currency();
+
+            Currency meso = new Currency(20000);
+            Currency meret = new Currency(2000);
+            Currency valorToken = new Currency(5);
+            Currency treva = new Currency(5);
+            Currency rue = new Currency(5);
+            Currency haviFruit = new Currency(5);
+            Currency mesoToken = new Currency(5);
```
AlanMorel, line 115: These all can just be done below

MapleServer2/Types/Player.cs
```diff
@@ -234,125 +278,116 @@ public void Warp(MapPlayerSpawn spawn, int mapId)
...
+                    return false;
                 default:
-                    break;
+                    return false;
             }
-            return true;
         }
 
-        public void SetCurrency(CurrencyType type, long amount)
+        public long GetCurrency(CurrencyType type)
```
AlanMorel, line 340: This should probably be called `GetCurrencyAmount` since it's not returning `Currency` it is returning the amount. That or make it return the currency and then callers just need to call `GetAmount()` on it

AlanMorel (changes requested):

MapleServer2/Enums/CurrencyType.cs
```diff
@@ -0,0 +1,31 @@
...
+    /* All currencies from wiki:
+     
+    Merets
+    Mesos
+    Havi Fruit
+    Rue(Currency)
+    Treva
+    Valor Token
+
+    Blue Star(Currency)
```
AlanMorel, line 23: I actually think it's harmless to just add these extra currencies as real enums

MapleServer2/Types/Player.cs
```diff
@@ -242,5 +251,126 @@ public void Warp(MapPlayerSpawn spawn, int mapId)
             Rotation = spawn.Rotation.ToFloat();
             Session.Send(FieldPacket.RequestEnter(Session.FieldPlayer));
         }
+
+
```
AlanMorel, line 257: Unneeded extra blank line

AlanMorel (changes requested):

MapleServer2/Enums/CurrencyType.cs
```diff
@@ -9,6 +9,13 @@ public enum CurrencyType
...
+        BlueStar,
+        RedStar,
+        EventDungeonCoin,
+        GuildCoins,
+        KayCoin,
+        MapleCoin,
+        PremiumCoin,
     }
 
     /* All currencies from wiki:
```
AlanMorel, line 21: Now we don't need this comment if all enums have been added

AlanMorel (changes requested):

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    public class Currency
+    {
+
+        public long Amount { get; private set; }
+        private readonly CurrencyType Type;
+        private readonly Player Player;
+
+        public Currency(CurrencyType type, long input, Player player)
+        {
```
AlanMorel, line 14: Can the order of the fields at the top, the parameters, and the initialization inside the constructor all be the same?

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,86 @@
+﻿using MapleServer2.Enums;
+using MapleServer2.Packets;
+
+namespace MapleServer2.Types
+{
+    public class Currency
+    {
+
```
AlanMorel, line 8: Unneeded extra blank line

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,86 @@
...
+                case CurrencyType.Meso:
+                    Player.Session.Send(MesosPacket.UpdateMesos(Player.Session));
+                    break;
+                case CurrencyType.Meret:
+                    Player.Session.Send(MeretsPacket.UpdateMerets(Player.Session));
+                    break;
+                case CurrencyType.GameMeret:
+                    Player.Session.Send(MeretsPacket.UpdateMerets(Player.Session));
+                    break;
+                case CurrencyType.EventMeret:
```
AlanMorel, line 54: All the cases that use the same packet can just be combined

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,86 @@
...
+                case CurrencyType.EventDungeonCoin:
+                    break;
+                case CurrencyType.GuildCoins:
+                    break;
+                case CurrencyType.KayCoin:
+                    break;
+                case CurrencyType.MapleCoin:
+                    break;
+                case CurrencyType.PremiumCoin:
+                    break;
```
AlanMorel, line 80: Likewise, for now we can just combine all these cases with a single `break`

MapleServer2/Types/Wallet.cs
```diff
@@ -0,0 +1,45 @@
+﻿using MapleServer2.Enums;
+
+namespace MapleServer2.Types
+{
+
```
AlanMorel, line 5: Unneeded extra blank line

AlanMorel (changes requested):

MapleServer2/Types/Currency.cs
bwu113: agreed. if all the wallet currencies use the same packet. combine them into 1 case and send that packet.

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,70 @@
...
+                case CurrencyType.Rue:
+                case CurrencyType.HaviFruit:
+                case CurrencyType.MesoToken:
+                case CurrencyType.BlueStar:
+                case CurrencyType.RedStar:
+                case CurrencyType.EventDungeonCoin:
+                case CurrencyType.GuildCoins:
+                case CurrencyType.KayCoin:
+                case CurrencyType.MapleCoin:
+                case CurrencyType.PremiumCoin:
```
rhydev, line 63: Any coin that is an item is not considered as a currency by the server, it is just an item so these should be removed.

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            if (input < 0)
+            {
+                return;
+            }
+            Amount = input;
+            UpdateUi();
+        }
+
+        private void UpdateUi()
```
rhydev, line 40: This function should be renamed, it's not just updating a UI value, it's sending data to the client. `UpdateClient` or similar would be more appropriate.

MapleServer2/Types/Currency.cs
```diff
@@ -0,0 +1,70 @@
...
+        public long Amount { get; private set; }
+
+        public Currency(Player player, CurrencyType type, long input)
+        {
+            Player = player;
+            Type = type;
+            Amount = input;
+        }
+
+        public bool Modify(long input)
```
rhydev, line 19: Seems like overflow is not handled here.

rhydev (changes requested):

MapleServer2/Types/Currency.cs
angelotadeucci: Do you know which ones are items?

MapleServer2/Types/Currency.cs
angelotadeucci: if (Amount + input < 0) is the overflow handling

MapleServer2/Types/Currency.cs
rhydev: All of the ones I put in the comment.

MapleServer2/Types/Currency.cs
rhydev: Ah right I forgot that overflow just turns negative.

MapleServer2/Types/Currency.cs
rhydev: iirc, the only non-item currencies are: mesos, merets, valor, treva, rue, havi fruit, and meso tokens.

PR #115, opened 2021-01-09
Fix: removed RedMeret declaration

angelotadeucci:


newgithubaccountlol (changes requested):
What is the reason for this change?

angelotadeucci:
> What is the reason for this change?

Red Merets don't exist anymore, it was modified to Event or Game Merets. And I forgot to removed on the main PR #107

PR #116, opened 2021-01-09
Feature(Wallet): Added packets

angelotadeucci:
Added packet for HaviFruit, Rue, Treva and Valor;
Added commands for testing purposes.

### DIFF
--- a/MapleServer2/Packets/WalletPacket.cs
+++ b/MapleServer2/Packets/WalletPacket.cs
@@ -1,0 +1,42 @@
+﻿using MaplePacketLib2.Tools;
+using MapleServer2.Constants;
+using MapleServer2.Enums;
+
+namespace MapleServer2.Packets
+{
+    class WalletPacket
+    {
+        public static Packet UpdateWallet(CurrencyType type, long amount)
+        {
+            byte currency = 0;
+            switch (type)
+            {
+                case CurrencyType.ValorToken:
+                    currency = 3;
+                    break;
+                case CurrencyType.Treva:
+                    currency = 4;
+                    break;
+                case CurrencyType.Rue:
+                    currency = 5;
+                    break;
+                case CurrencyType.HaviFruit:
+                    currency = 6;
+                    break;
+                default:
+                    break;
+            }
+
+            PacketWriter pWriter = PacketWriter.Of(SendOp.MONEY_TOKEN);
+
+            pWriter.WriteByte(currency); // type of currency
+            pWriter.WriteLong(amount); // currency amount
+            pWriter.Write("FF FF FF FF FF FF FF FF".ToByteArray()); // always the same
+            pWriter.WriteShort(52); // always the same
+            pWriter.WriteLong(); // unk
+            pWriter.WriteShort(); // unk
+
+            return pWriter;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


MapleServer2/Packets/WalletPacket.cs
```diff
@@ -0,0 +1,42 @@
...
+                    break;
+                default:
+                    break;
+            }
+
+            PacketWriter pWriter = PacketWriter.Of(SendOp.MONEY_TOKEN);
+
+            pWriter.WriteByte(currency); // type of currency
+            pWriter.WriteLong(amount); // currency amount
+            pWriter.Write("FF FF FF FF FF FF FF FF".ToByteArray()); // always the same
```
newgithubaccountlol, line 34: WriteLong(-1)

MapleServer2/Packets/WalletPacket.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    class WalletPacket
+    {
+        public static Packet UpdateWallet(CurrencyType type, long amount)
+        {
+            byte currency = 0;
+            switch (type)
+            {
+                case CurrencyType.ValorToken:
+                    currency = 3;
```
AlanMorel, line 15: Is it possible to just give these currencies an enum value? That should let you not need this switch case at all.

MapleServer2/Tools/GameCommandActions.cs
```diff
@@ -14,9 +14,21 @@ public static class GameCommandActions
     {
         public static void Process(GameSession session, string command)
         {
-            string[] args = command.ToLower().Split(" ", 2);
+            string[] args = command.ToLower().Split(" ");
             switch (args[0])
             {
+                case "setvalor":
+                    session.Player.Wallet.ValorToken.SetAmount(ParseLong(session, args.Length > 1 ? args[1] : ""));
```
newgithubaccountlol, line 21: this code is repeated many times. it seems like it's doing it for all cases so you can just move it out.

AlanMorel (changes requested):

MapleServer2/Packets/WalletPacket.cs
newgithubaccountlol, line 15: do we want to add currency type to the enum?
https://www.tutorialsteacher.com/csharp/csharp-enum

newgithubaccountlol (changes requested):

MapleServer2/Packets/WalletPacket.cs
angelotadeucci: Sure!

MapleServer2/Tools/GameCommandActions.cs
angelotadeucci: Sorry, you mean remove the command?

AlanMorel: You can't move it out because it's calling it on a different currency each case

AlanMorel (changes requested):

MapleServer2/Tools/GameCommandActions.cs
newgithubaccountlol: long amount = ParseLong(session, args.Length > 1 ? args[1] : ""));
...
...SetAmount(amount);

newgithubaccountlol: Never mind, I'm wrong about how these commands work.

PR #117, opened 2021-01-10
DB Character (With minor Account changes)

bwu113:
Missing Inventory slots, Equipment slots, Skill tree, etc. 
Single value variables are all implemented. (level, jobid, name, etc.)

### DIFF
--- a/Ms2Database/Controllers/AccountManager.cs
+++ b/Ms2Database/Controllers/AccountManager.cs
@@ -11,7 +11,7 @@ namespace Ms2Database.Controllers
 {
     public class AccountManager
     {
-        public void CreateAccount(string username, string password) // Creates account entry
+        public void CreateAccount(string username, string password)
         {
             using (Ms2DbContext context = new Ms2DbContext())
             {
@@ -31,10 +31,6 @@ namespace Ms2Database.Controllers
             using (Ms2DbContext context = new Ms2DbContext())
             {
                 Account account = context.Accounts.FirstOrDefault(a => a.AccountId == id); // Retrieve entry by Id
-                if (account == null) // Checks if accountId exists (Possibly just make a error box appear instead of returning null)
-                {
-                    return null;
-                }
                 return account;
             }
         }
--- a/Ms2Database/Controllers/AccountManager.cs
+++ b/Ms2Database/Controllers/AccountManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Ms2Database.DbClasses;
 using System.Data.Entity;
+using System.Diagnostics;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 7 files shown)


Ms2Database/Controllers/AccountManager.cs
```diff
@@ -10,20 +11,44 @@ namespace Ms2Database.Controllers
 {
     public class AccountManager
     {
-        public Tuple<string, string> GetAccInfoById(long id) // Queries db and retrieves account entry by Id
+        public void CreateAccount(string username, string password) // Creates account entry
```
AlanMorel, line 14: Comment isn't needed, seems self-explanatory

Ms2Database/Controllers/AccountManager.cs
```diff
@@ -10,20 +11,44 @@ namespace Ms2Database.Controllers
...
+            }
+        }
+
+        public Account GetAccInfoById(long id) // Queries db and retrieves account entry by Id
         {
             using (Ms2DbContext context = new Ms2DbContext())
             {
                 Account account = context.Accounts.FirstOrDefault(a => a.AccountId == id); // Retrieve entry by Id
-                return new Tuple<string, string>(account.Username, account.Password);
+                if (account == null) // Checks if accountId exists (Possibly just make a error box appear instead of returning null)
```
AlanMorel, line 34: If it's null, it will be returned as such on line 38

Ms2Database/Controllers/CharacterManager.cs
```diff
@@ -0,0 +1,56 @@
...
+        public Character GetCharInfo(long charId)
+        {
+            using (Ms2DbContext context = new Ms2DbContext())
+            {
+                Character character = context.Characters.FirstOrDefault(a => a.CharacterId == charId);
+                return character;
+            }
+        }
+
+        public void SetCharInfo(Character CharIn)
```
AlanMorel, line 47: `charIn`

AlanMorel (changes requested):

Ms2Database/Controllers/CharacterManager.cs
bwu113: what do u not understanding what u mean here.
Idea is to have whatever is calling it. pass in the character object.

bwu113: or its too many parameters to pass in.

bwu113: i'm good for either one just think this is a bit cleaner

newgithubaccountlol (commented):
I don't see anything particularly wrong. I think this is good. How was this tested though?

bwu113:
This was tested with a test data entry and by viewing table structure via database after creation. As well as test cases of when search functions are empty or wrong.

Ms2Database/DBClasses/Ms2DBContext.cs
```diff
@@ -11,8 +11,10 @@ public class Ms2DbContext : DbContext
     {
         public Ms2DbContext() : base("name=DatabaseConnectionString") // Creates context for database via connection string
         {
-            //Database.SetInitializer<Ms2DBContext>(new DropCreateDatabaseAlways<Ms2DBContext>());
+            //Database.SetInitializer<Ms2DbContext>(new DropCreateDatabaseAlways<Ms2DbContext>());
```
eotetb, line 14: Is this just temporarily commented out? May want to add a TODO comment just in case.

Ms2Database/DbClasses/Character.cs
```diff
@@ -0,0 +1,113 @@
...
+        public long MesoToken { get; set; }
+
+        public long? PartyId { get; set; }
+
+        public long? ClubId { get; set; }
+
+        public long? GuildId { get; set; }
+
+        [StringLength(20)]
+        public string GuildeName { get; set; }
```
eotetb, line 90: Typo for GuildName?

Ms2Database/DbClasses/Character.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public string ProfileUrl { get; set; }
```
eotetb, line 92: I recommend setting a maximum string length equal to whatever you have it set to in the database. Since it is a url, it probably should be fairly large.

Ms2Database/Program.cs
```diff
@@ -4,25 +4,22 @@
...
-            {
-                Account account = new Account()
-                {
-                    Username = "localhost",
-                    Password = ""
-                };
-                context.Accounts.Add(account);
-                context.SaveChanges();
-            }
+            AccountManager accmanage = new AccountManager();
```
eotetb, line 15: Variable style according to coding style guidelines, use `accManage` or `accountManager`. I personally don't recommend abbreviating, it can make some things harder to understand, although it is pretty clear in this case.

eotetb (commented):
Left a few comments. Good overall, its mostly minor nitpicks.

Ms2Database/DBClasses/Ms2DBContext.cs
bwu113: This is temporarily commented out. It basically recreates the database on each run. Which I use consistently to test database structuring. Once The whole database is finished structuring. This line of code will be gone.

Ms2Database/DbClasses/Character.cs
bwu113: Oh thanks I didn't see that lol

Ms2Database/DbClasses/Character.cs
bwu113: Its automatically set to max length in database.

newgithubaccountlol (approved):

PR #118, opened 2021-01-10
Separating Mob from Npc, Handle Damage & Stat for Mobs

Sparkymod:
- Passing a list of mobs from the field to the SkillDamagePacket
- Adding `AddMob, LoadMob & ControlMob` to be use for the Mob class
- Adding `ApplyHeal` to the Packet
- Adding `MobSkillUse` to the Packet
- Adding `StatPacket` into the proper Packet class
- Adding `mob` calling on `GameCommandAction`
- Adding spawing mobs to into the state of the field.
- Minor other fix

### DIFF
--- a/MapleServer2/PacketHandlers/Game/SkillHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/SkillHandler.cs
@@ -30,10 +30,6 @@ namespace MapleServer2.PacketHandlers.Game
             TypeOfDamage2 = 0x2
         }
 
-        private long skillUid;
-        private int skillId;
-        private short skillLevel;
-
         public override void Handle(GameSession session, PacketReader packet)
         {
             SkillHandlerMode mode = (SkillHandlerMode) packet.ReadByte();
@@ -59,14 +55,14 @@ namespace MapleServer2.PacketHandlers.Game
 
         private void HandleFirstSent(GameSession session, PacketReader packet)
         {
-            skillUid = packet.ReadLong();
+            long skillUid = packet.ReadLong();
             int value = packet.ReadInt();
-            skillId = packet.ReadInt();
-            skillLevel = packet.ReadShort();
+            int skillId = packet.ReadInt();
+            short skillLevel = packet.ReadShort();
             packet.ReadByte();
             CoordF coords = packet.Read<CoordF>();
             packet.ReadShort();
-            session.Send(SkillUsePacket.SkillUse(value, skillUid, skillId, skillLevel, coords));
+            session.Send(SkillUsePacket.SkillUse(session.FieldPlayer, value, skillUid, coords));
         }
 
         private void HandleDamage(GameSession session, PacketReader packet)
@@ -102,7 +98,7 @@ namespace MapleServer2.PacketHandlers.Game
 
         private void HandleTypeOfDamage(GameSession session, PacketReader packet)
         {
-            skillUid = packet.ReadLong();
+            long skillUid = packet.ReadLong();
             packet.ReadByte();
             CoordF coords = packet.Read<CoordF>();
             CoordF coords2 = packet.Read<CoordF>();
@@ -123,7 +119,7 @@ namespace MapleServer2.PacketHandlers.Game
         private void HandleAoeDamage(GameSession session, PacketReader packet)
         {
             List<IFieldObject<Mob>> mobs = new List<IFieldObject<Mob>>();
-            skillUid = packet.ReadLong();
+            long skillUid = packet.ReadLong();
             int someValue = packet.ReadInt();
             int playerObjectId = packet.ReadInt();
             CoordF coords = packet.Read<CoordF>();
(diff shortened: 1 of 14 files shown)


MapleServer2/Packets/StatPacket.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.STAT);
+            pWriter.WriteInt(mob.ObjectId);
+            pWriter.WriteByte();
+            pWriter.WriteByte(1);
+            pWriter.WriteByte(4); // value
+            // Stats 
+            pWriter.WriteLong(mob.Value.stats.Hp.Total);
+            pWriter.WriteLong(mob.Value.stats.Hp.Max);
+            pWriter.WriteLong(mob.Value.stats.CurrentHp.Min -= damage);
```
AlanMorel, line 28: Manipulation of mob stats should not happen in this method

AlanMorel (changes requested):

MapleServer2/Packets/StatPacket.cs
Sparkymod: Should be in another class like `Damage` and call a method with the new stats value?

MapleServer2/PacketHandlers/Game/SkillHandler.cs
```diff
@@ -53,14 +59,14 @@ public override void Handle(GameSession session, PacketReader packet)
 
         private void HandleFirstSent(GameSession session, PacketReader packet)
         {
-            long skillCount = packet.ReadLong();
```
newgithubaccountlol, line 56: As stated above, keep these local variables. It makes no sense to set them elsewhere

MapleServer2/Packets/StatPacket.cs
newgithubaccountlol: Whatever calls this method should be doing the mutation

MapleServer2/PacketHandlers/Game/SkillHandler.cs
Sparkymod: declaring it global in the class but private, give me the option to pass the value that needs from one method to another because is wip right now until skill are handle the correct way, what do you suggest? probably make another class to handle this variables to use it in the methods i need? for example, `skillId, skillLevel` is read in `HandleFirstSent` and i need them on `HandleAoeDamage` but is not being send on that packet.

MapleServer2/Packets/StatPacket.cs
Sparkymod: Sure!

MapleServer2/Packets/FieldObjectPacket.cs
```diff
@@ -90,6 +90,17 @@ public static Packet LoadNpc(IFieldObject<Npc> npc)
...
 
+        public static Packet LoadMob(IFieldObject<Mob> mob)
+        {
+            return PacketWriter.Of(SendOp.PROXY_GAME_OBJ)
+                .WriteByte(0x6)
+                .WriteInt(mob.ObjectId)
+                .WriteInt(mob.Value.Id)
+                .WriteByte()
+                .WriteInt(200)
+                .Write(mob.Coord);
```
rhydev, line 101: Prefer non-chaining method calls for writing packets, makes it easier to organize and make changes.

MapleServer2/Packets/FieldObjectPacket.cs
```diff
@@ -110,6 +121,26 @@ public static Packet ControlNpc(IFieldObject<Npc> npc)
...
+            PacketWriter npcBuffer = new PacketWriter()
+                .WriteInt(mob.ObjectId)
+                .WriteByte()
+                .Write(mob.Coord.ToShort())
+                .WriteShort(mob.Value.Rotation)
+                .Write(mob.Value.Speed) // XYZ Speed
+                .WriteShort(100) // Unknown
+                .WriteByte(1) // Flag ?
+                .WriteShort(mob.Value.Animation)
+                .WriteShort(1); // counter (increments every packet)
```
rhydev, line 135: Again prefer non-chaining for packet writing.

MapleServer2/Packets/FieldObjectPacket.cs
```diff
@@ -110,6 +121,26 @@ public static Packet ControlNpc(IFieldObject<Npc> npc)
...
+            // There can be more to this packet, probably dependent on Flag.
+
+            return PacketWriter.Of(SendOp.NPC_CONTROL)
+                .WriteShort(1) // Segments
+                .WriteShort((short) npcBuffer.Length)
+                .Write(npcBuffer.ToArray());
```
rhydev, line 141: Again prefer non-chaining for packet writing.

rhydev (changes requested):

MapleServer2/Packets/FieldObjectPacket.cs
```diff
@@ -11,25 +11,25 @@ public static class FieldObjectPacket
...
-                .WriteShort((short) player.JobGroupId)
-                .WriteInt(player.JobId)
-                .WriteInt()
-                .WriteInt()
-                .WriteInt()
-                .WriteUnicodeString(player.HomeName)
-                .WriteInt()
-                .WriteShort();
+            PacketWriter pWriter = PacketWriter.Of(SendOp.PROXY_GAME_OBJ);
+            pWriter.WriteByte(0x03);
```
rhydev, line 15: I'm assuming this is a mode, can we handle modes properly? (mode enum, `.WriteMode`)

rhydev (approved):

Issue #83, opened 2020-12-30
Guilds


integralcoder95:
- Creating
- Joining
- Viewing members
- Searching for guilds
- Guild applications
- Setting guild ranks
- Creating custom guild roles
- Guild trophies
- Guild house
- Guilds level
- Guild NPCs
- Guild raids

rhydev:
https://github.com/AlanMorel/MapleServer2/pull/100

Sparkymod:
#143

PR #119, opened 2021-01-10
Guild packets for guild listing, applying, rank changing, buffs, and services

Zintixx:
I'm calling these initial features or just adding in the packets because they're not fully functioning. and I'll most likely need either someone else to assist on it or implement it
Guild listing is not working as intended due to it needing to grab a list of the guilds there are.
Applying is not working as intended because this will need some way to store applications with the respective player's info
Rank changing is not working as intended because this will also need to be stored for each member
Buff packets are there but will need buffs to be fully functioning for this to work
Hiring NPC services are sort of added. Need to properly implement a way to update a guild with it. Once guild housing is implemented, this can be used for it.

#83

### DIFF
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -1,10 +1,14 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
+using System;
 using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
+using MapleServer2.Enums;
 using MapleServer2.Packets;
 using MapleServer2.Servers.Game;
-using Microsoft.Extensions.Logging;
+using MapleServer2.Tools;
 using MapleServer2.Types;
+using Microsoft.Extensions.Logging;
 
 namespace MapleServer2.PacketHandlers.Game
 {
@@ -22,15 +26,20 @@ namespace MapleServer2.PacketHandlers.Game
             InviteResponse = 0x5,
             Leave = 0x7,
             Kick = 0x8,
+            RankChange = 0xA,
             PlayerMessage = 0xD,
             CheckIn = 0xF,
             TransferLeader = 0x3D,
             GuildNotice = 0x3E,
             ListGuild = 0x42,
-            GuildWindow = 0x54,
-            List = 0x55,
+            SubmitApplication = 0x51,
+            ApplicationResponse = 0x53,
+            GuildWindowRequest = 0x54,
+            GuildListRequest = 0x55,
+            UseBuff = 0x59,
             EnterHouse = 0x64,
             GuildDonate = 0x6E,
+            Services = 0x6F,
         }
 
         public override void Handle(GameSession session, PacketReader packet)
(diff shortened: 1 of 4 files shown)


newgithubaccountlol (approved):
Looks good to me

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -93,25 +117,27 @@ private void HandleCreate(GameSession session, PacketReader packet)
         {
             string guildName = packet.ReadUnicodeString();
 
-            if (session.Player.Wallet.Meso.Modify(-2000))
+            if (session.Player.Wallet.Meso.Amount < 2000)
```
angelotadeucci, line 120: This is the wrong way of using function Modify, it returns an boolean, if the character doesn't have enough money it returns false and if it does have enough money the function changes the amount and returns true.

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -324,14 +368,69 @@ private void HandleListGuild(GameSession session, PacketReader packet)
...
+            List<Guild> guildList = GameServer.GuildManager.GetGuildList();
+            guildList = guildList.OrderBy(g => g.Members.Count).ToList();
+            session.Send(GuildPacket.DisplayGuildList(guildList));
+        }
+
+        private void HandleUseBuff(GameSession session, PacketReader packet)
+        {
+            int buffID = packet.ReadInt();
+
+            if (session.Player.Wallet.Meso.Amount < 100000)
```
angelotadeucci, line 426: Same thing here (read my other comment) you don't need to access Amount if Modify already does the check for you.

angelotadeucci (changes requested):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
Zintixx: What was there previously did not work. It would take your mesos and tell you you don't have enough mesos.

angelotadeucci: It's because when I made the Wallet class I didn't change anything else outside the currency feat, so it's my bad, the IF should be inverted then it will work without a problem. Sorry!

angelotadeucci (commented):
Looks good!

angelotadeucci (approved):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -117,15 +117,14 @@ private void HandleCreate(GameSession session, PacketReader packet)
...
             string guildName = packet.ReadUnicodeString();
 
-            if (session.Player.Wallet.Meso.Amount < 2000)
+            if (!session.Player.Wallet.Meso.Modify(-2000))
             {
                 short NoticeCode = 5121;
                 session.Send(GuildPacket.ErrorNotice(NoticeCode));
                 return;
             }
             else
```
AlanMorel: Since you're returning, you can safely remove this else and unindent the following code

AlanMorel (changes requested):

angelotadeucci (approved):

PR #124, opened 2021-01-13
Fix: force en-US globalization

rhydev:
`type.Parse()` uses OS globalization to determine whether to use `,` or `.` for decimal points, this change forces it to use `.` for decimal points.


angelotadeucci:
I did this. :)

angelotadeucci (approved):

PR #125, opened 2021-01-13
Stickers

Zintixx:


angelotadeucci (approved):

AlanMorel (approved):

rhydev (changes requested):
Can we name the handlers and packets after the opcode similar to all other packets and handlers?

rhydev (approved):

PR #127, opened 2021-01-14
Fix: Refactoring Skills

Sparkymod:
- Using SkillMetadata instead of each `skills.xml`
- Updating Job to use SkillMetadata.
- Removing `skills.xml` files


rhydev (changes requested):
Where is sub skill handling? Also where are the default skills for wall climb etc.? We likely need to parse different skill files.

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -44,69 +43,57 @@ public static List<SkillMetadata> Parse(MemoryMappedFile m2dFile, IEnumerable<Pa
                     }
                     XmlNode kinds = node.SelectSingleNode("kinds");
                     metadata.Type = kinds.Attributes["type"].Value != null ? int.Parse(kinds.Attributes["type"].Value) : 0;
-                    metadata.SubType = kinds.Attributes["subType"].Value != null ? int.Parse(kinds.Attributes["subType"].Value) : 0;
-                    metadata.RangeType = kinds.Attributes["rangeType"].Value != null ? int.Parse(kinds.Attributes["rangeType"].Value) : 0;
-                    metadata.Element = kinds.Attributes["element"].Value != null ? int.Parse(kinds.Attributes["element"].Value) : 0;
+                    if (kinds.Attributes.GetNamedItem("subType") != null)
+                    {
+                        // Get Rank
+                        metadata.SubType = kinds.Attributes["subType"].Value != null ? int.Parse(kinds.Attributes["subType"].Value) : 0;
```
newgithubaccountlol, line 49: redundant check. also if it's null, then SubType is not being set (don't think that matters though?)

MapleServer2/Types/Player.cs
```diff
@@ -109,12 +106,11 @@ public Player()
         public static Player Char1(long accountId, long characterId, string name = "Char1")
         {
             int job = 50; // Archer
-
             PlayerStats stats = PlayerStats.Default();
             StatDistribution StatPointDistribution = new StatDistribution();
             List<SkillTab> skillTabs = new List<SkillTab>
             {
-                XmlParser.ParseSkills(job)
+                new SkillTab((Job) (job*10))
```
newgithubaccountlol, line 113: should we make constants for this?

MapleServer2/Enums/Job.cs
```diff
@@ -17,22 +17,38 @@ public enum Job
...
-        Wizard = 301,
-        Priest = 401,
-        Archer = 501,
-        HeavyGunner = 601,
-        Thief = 701,
-        Assassin = 801,
-        Runeblade = 901,
-        Striker = 1001,
-        SoulBinder = 1101,
+        Berserker = 102,
```
newgithubaccountlol, line 24: this way of coding it and where you add 400 to the job id in SkillMetaData seems very confusing even though the ids seem to work. If anything this should be labelled only for skills

MapleServer2/Data/Static/SkillMetadataStorage.cs
```diff
@@ -1,28 +1,63 @@
...
+        public static SkillMetadata GetSkill(int id)
         {
-            return skillLevel.GetValueOrDefault(level);
+            return skill.GetValueOrDefault(id);
+        }
+
+        public static List<SkillMetadata> GetJobSkills(Job job = Job.None)
+        {
+            List<SkillMetadata> jobSkill = new List<SkillMetadata>();
+            int jobCode = ((Convert.ToInt32(job) / 100) + 100) * 100000; // 10X0000 10100000
```
newgithubaccountlol, line 33: Probably should specify the skill ids are in this format because otherwise it's not obvious at all

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        /// <summary>
+        /// Get the specific Split value for each Job. Requiered for SkillBookTree
+        /// </summary>
+        public static byte GetListOrderedSplit(Job job)
+        {
+            switch (job)
+            {
+                case Job.Knight:
+                    return (byte) JobSkillSplit.Knight;
```
newgithubaccountlol, line 206: any reason why these are bytes?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -44,69 +43,57 @@ public static List<SkillMetadata> Parse(MemoryMappedFile m2dFile, IEnumerable<Pa
...
+                        // Get Rank
+                        metadata.DefaultSkill = stateAttr.Attributes["useDefaultSkill"].Value != null ? byte.Parse(stateAttr.Attributes["useDefaultSkill"].Value) : 0;
+                    }
                 }
 
                 XmlNodeList levels = document.SelectNodes("/ms2/level");
                 foreach (XmlNode node in levels)
                 {
-                    int level = node.Attributes["value"].Value != null ? int.Parse(node.Attributes["value"].Value) : 1;
-                    XmlNode consume = node.SelectSingleNode("consume/stat");
```
newgithubaccountlol, line 56: are we still using xml? I see them all deleted

newgithubaccountlol (changes requested):

GameDataParser/Parsers/SkillParser.cs
Sparkymod: Alright, just leaving the lambda for the check.

MapleServer2/Types/Player.cs
Sparkymod: Sure! changing on player the `int Job` to the `Job job` and using it on the Packet for character too

MapleServer2/Enums/Job.cs
Sparkymod: True, it was for testing, ill change it on the next commit

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
Sparkymod: Is because on the `SkillTreePacket` is being pass as byte

GameDataParser/Parsers/SkillParser.cs
Sparkymod: Yes, they are being deleted, however, we are parsing skill from files and using it on a metadata file, the same way as `map` or `item`

MapleServer2/Data/Static/SkillMetadataStorage.cs
Sparkymod: Didn't get what you mean on this, could you elaborate more ?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
 {
     public static class SkillParser
     {
-
         public static List<SkillMetadata> Parse(MemoryMappedFile m2dFile, IEnumerable<PackFileEntry> entries)
         {
             List<SkillMetadata> skillList = new List<SkillMetadata>();
-
-            foreach (PackFileEntry skill in entries)
+            foreach (PackFileEntry xmlFile in entries)
```
rhydev, line 20: Why are we calling it `xmlFile` instead of `entry` like every other parser?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
-                metadata.SkillId = int.Parse(skillId);
-                Debug.Assert(metadata.SkillId > 0, $"Invalid Id {metadata.SkillId} from {skillId}");
-
-                //using XmlReader reader = m2dFile.GetReader(skill.FileHeader);
-                XmlDocument document = m2dFile.GetDocument(skill.FileHeader);
-
-                XmlNodeList basic = document.SelectNodes("/ms2/basic");
-                foreach (XmlNode node in basic)
+                // Parsing Skills
+                if (xmlFile.Name.StartsWith("skill"))
```
rhydev, line 23: Invert and use `continue`

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
-                {
-                    int level = node.Attributes["value"].Value != null ? int.Parse(node.Attributes["value"].Value) : 1;
-                    XmlNode consume = node.SelectSingleNode("consume/stat");
-
-                    int spirit = 0;
-                    int upgradeLevel = 0;
-                    int[] upgradeSkillId = new int[0];
-                    int[] upgradeSkillLevel = new int[0];
+                    metadata.SkillId = int.Parse(skillId);
+                    Debug.Assert(metadata.SkillId > 0, $"Invalid Id {metadata.SkillId} from {skillId}");
```
rhydev, line 30: Seems like this isn't supposed to be here

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
```diff
@@ -0,0 +1,271 @@
...
+        #endregion
+        /// <summary>
+        /// Get the specific ordered skill list of each Job. Requiered for SkillBookTree
+        /// </summary>
+        public static List<int> GetListOrdered(Job job)
+        {
+            switch (job)
+            {
+                case Job.Knight:
+                    return Knight;
```
AlanMorel, line 169: If you turn the above code into a Dictionary, where the key is the Job and the values is the list of skills, you can eliminate the need for this method and clean up the above code

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
-                    if (consume.Attributes.GetNamedItem("sp") != null)
-                    {
-                        spirit = int.Parse(consume.Attributes["sp"].Value);
-                    }
+                    XmlDocument document = m2dFile.GetDocument(xmlFile.FileHeader);
 
-                    XmlNode upgrade = node.SelectSingleNode("upgrade");
-                    if (upgrade.Attributes != null)
+                    XmlNodeList basic = document.SelectNodes("/ms2/basic");
+                    foreach (XmlNode node in basic)
```
rhydev, line 35: Why are we using a loop here when we know there is only 1 `basic` element?

AlanMorel (changes requested):

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
                     {
-                        if (upgrade.Attributes.GetNamedItem("level") != null)
-                        {
-                            upgradeLevel = int.Parse(upgrade.Attributes["level"].Value);
-                        }
-                        if (upgrade.Attributes.GetNamedItem("skillIDs") != null)
+                        if (node.Attributes.GetNamedItem("feature") != null)
```
rhydev, line 37: What is the use of parsing the feature here? Is it necessary?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
 
-                    XmlNode motion = node.SelectSingleNode("motion/motionProperty");
-
-                    string sequenceName = "";
-                    string motionEffect = "";
-                    string strTagEffects = "";
-
-                    if (motion.Attributes != null)
+                    XmlNodeList levels = document.SelectNodes("/ms2/level");
+                    foreach (XmlNode node in levels)
```
rhydev, line 45: Please follow good practice and when iterating over lists use the singular name of the list so `XmlNode level in levels`

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
                     {
-                        if (motion.Attributes.GetNamedItem("sequenceName") != null)
+                        if (node.Attributes.GetNamedItem("feature") == null || node.Attributes.GetNamedItem("feature").Value == metadata.Feature)
```
rhydev, line 47: What is this for?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
                         {
-                            sequenceName = motion.Attributes["sequenceName"].Value;
+                            int level = node.Attributes["value"].Value != null ? int.Parse(node.Attributes["value"].Value) : 1;
```
rhydev, line 49: Is there a reason we're defaulting this to 1 here, what is the use case?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
+                            XmlNode consume = node.SelectSingleNode("consume/stat");
+
+                            int spirit = 0;
+                            int upgradeLevel = 0;
+                            int[] upgradeSkillId = new int[0];
+                            int[] upgradeSkillLevel = new int[0];
+
+                            if (consume.Attributes.GetNamedItem("sp") != null)
```
rhydev, line 57: Is there a reason we're using `GetNamedItem` instead of just generic square bracket indexing ie. `Attributes["sp"]`?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
+                            {
+                                spirit = int.Parse(consume.Attributes["sp"].Value);
+                            }
+
+                            XmlNode upgrade = node.SelectSingleNode("upgrade");
```
rhydev, line 62: What is this for?

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
-                        if (motion.Attributes.GetNamedItem("motionEffect") != null)
+                    }
+                    skillList.Add(metadata);
+                }
+                // Parsing SubSkills
+                else if (xmlFile.Name.StartsWith("table/job"))
+                {
+                    XmlDocument document = m2dFile.GetDocument(xmlFile.FileHeader);
+                    XmlNodeList job = document.SelectNodes("/ms2/job");
+                    foreach (XmlNode node in job)
```
rhydev, line 89: Again let's use proper variable naming `XmlNode job in jobs`

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
+                                            {
+                                                if (skillList.Select(x => x.SkillId).Contains(sub[n]))
+                                                {
+                                                    skillList.Find(x => x.SkillId == sub[n]).Job = jobCode;
+                                                }
+                                            }
+                                        }
+                                    }
+                                }
+                            }
```
rhydev, line 120: What in the world is this doing? The sub skills just need to be an array of skill ids

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -13,101 +14,127 @@ namespace GameDataParser.Parsers
...
                         }
-                        if (motion.Attributes.GetNamedItem("strTagEffects") != null)
+                        else if (node.Attributes.GetNamedItem("code").Value == "001")
```
rhydev, line 122: Why are we using `GetNamedItem`

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
```diff
@@ -0,0 +1,271 @@
+﻿using System.Collections.Generic;
+using MapleServer2.Enums;
+
+namespace MapleServer2.Constants.Skills
+{
+    public static class SkillTreeOrdered
```
rhydev, line 6: What happened to being able to send the skills in the order they are listed in the xml? Did it end up not working?

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
```diff
@@ -0,0 +1,271 @@
...
+            11100051,11100104,11100105,11100031,11100321,11100204,11100187,11100272,11100001,11100291,
+            11100002,11100173,11100243,11100211,11100262,11100161,11100301,11100063,11100251
+        };
+        private static readonly List<int> GameMaster = new List<int>
+        {
+            0
+        };
+        #endregion
+
+        #region Default skills learned.
```
rhydev, line 115: These default skills are literally in the xml

MapleServer2/Packets/SkillBookTreePacket.cs
```diff
@@ -20,16 +21,17 @@ public static Packet Open(Player character)
             pWriter.WriteUnicodeString("Build 1"); // Page name
 
             // Get feature skills, for now just rank 1 skills, not sure how to tell which tab is opened
-            List<Skill> skills = character.SkillTabs[0].GetJobFeatureSkills(); // Get first skill tab skills only for now, uncertain of how to have multiple skill tabs
+            // Get first skill tab skills only for now, uncertain of how to have multiple skill tabs
+            List<SkillMetadata> skills = character.SkillTabs[0].GetJobFeatureSkills((Enums.Job) character.JobGroupId);
             skills.RemoveAll(x => x.Learned < 1); // Remove all unlearned skills
-            skills.RemoveAll(x => x.Id == 10900051 || x.Id == 10900041); // temp remove rb default passives
+            skills.RemoveAll(x => x.SkillId == 10900051 || x.SkillId == 10900041); // temp remove rb default passives
```
rhydev, line 27: This should probably be removed

Ms2Database/App.config
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
rhydev, line 1: What is this file doing here?

rhydev (changes requested):

GameDataParser/Parsers/SkillParser.cs
Sparkymod: cannot be because it then won't get the other parse on 'table/job'

GameDataParser/Parsers/SkillParser.cs
Sparkymod: i renamed to `if (level.Attributes["feature"] == null` and this is use filter the others levels under the `feature=CharacterBalance` and not all skills beside `Job` has it.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: This is getting all the Skill that contains a subSkill and adding in into the corresponding skill, and because is from another xml in the same file is getting filtered under `feature=JobChange_02` that contains all the list, same as it was doing with `string/skill_{}` before.

Didn't get a better solution so far.

MapleServer2/Constants/Skills/SkillTreeOrdered.cs
Sparkymod: it wasn't loading the second passive skill that all jobs have from default and didn't get the same result with using the `table/job` so instead i load the subskill list for each job that they are all there and in future use the `feature` and `subJobCode` to define when it is awakened or not and load the corresponding skill.

Ms2Database/App.config
Sparkymod: How do i remove this ? i think it got here because the rebase that ocurre this days.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: This are the skill requiered for the skill lvl up. Can this be removed then ?

GameDataParser/Parsers/SkillParser.cs
rhydev: Yes you can, it's called `||`...

GameDataParser/Parsers/SkillParser.cs
rhydev: Still doesn't answer the question of why it exists, are you saying you don't want to parse all levels?

GameDataParser/Parsers/SkillParser.cs
rhydev: As in prerequisite skill before a skill can be learned?

Ms2Database/App.config
rhydev: Remove the file

GameDataParser/Parsers/SkillParser.cs
rhydev: The skills in table/job have an attribute with the corresponding sub skills, we can just pull the ids from there and put them into an array.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: indeed, but seems that the client handles that on its own.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: `sub = Array.ConvertAll(skills.ChildNodes[i].Attributes["sub"].Value.Split(","), int.Parse);`
That is what im doing there, then im adding on each skill the sub it correspond with `skill.SubSkill = sub;`
And because they also are skills from the same `code` i need to go through each subskill and add them the jobCode as well.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: Now i remember, It was default to 1 because all other skills beside Jobs has just 1 level and to default them all instead of parsing the data and iterates on all, wich we not be using now, default them to 1 will work.

GameDataParser/Parsers/SkillParser.cs
Sparkymod: All skills have just 1 level, besides jobs, why i would parse all those other, wich i wont be using but just need the lvl? instead i default them to 1 and filter those who has more than 1 levels

GameDataParser/Parsers/SkillParser.cs
rhydev: Yeah but I'm saying we don't need to have anything other than the id, once you have the id and we want to level those skills up we can just go through the subskill array, the sub skills should all already be in the skill array, they should all already be defined, unless they aren't?

GameDataParser/Parsers/SkillParser.cs
rhydev: Can you give an example of some non-job skills?

GameDataParser/Parsers/SkillParser.cs
Sparkymod: They aren't define on `skills` they are define either on `table/job` or `string/skills{number}` and `table/job` has them all and the default ones, we just grab the subskills and the defaults skill from that and added into the list we already have.

Sparkymod: And this is doing the check of the id for each job and adding the `jobCode` to that skill, and then is adding the `sub skill array` to the skill corresponding, and then on each `subskill` is adding the `jobCode` as well

Maple2Storage/Types/Metadata/SkillMetadata.cs
```diff
@@ -89,121 +41,55 @@ public class SkillLevel
...
-            this.Level = _Level;
-            this.Spirit = _Spirit;
-            this.UpgradeLevel = _UpgradeLevel;
-            this.UpgradeSkillId = _UpgradeSkillId;
-            this.UpgradeSkillLevel = _UpgradeSkillLevel;
-            this.SkillMotion = motion;
         }
 
-        protected bool Equals(SkillLevel other)
+        public SkillLevel(int _Level, int _Spirit, float _DamageRate, string _Feature)
```
rhydev, line 54: Use camelcase for parameter variables

Maple2Storage/Types/Metadata/SkillMetadata.cs
```diff
@@ -11,74 +11,26 @@ public class SkillMetadata
         [XmlElement(Order = 1)]
         public int SkillId;
         [XmlElement(Order = 2)]
-        public string Feature = "";
+        public List<SkillLevel> SkillLevel;
         [XmlElement(Order = 3)]
-        public int AttackType;
+        public int[] SubSkill = new int[0];
```
rhydev, line 16: Arrays and lists should be plural: `SubSkills`

Maple2Storage/Types/Metadata/SkillMetadata.cs
```diff
@@ -11,74 +11,26 @@ public class SkillMetadata
         [XmlElement(Order = 1)]
         public int SkillId;
         [XmlElement(Order = 2)]
-        public string Feature = "";
+        public List<SkillLevel> SkillLevel;
```
rhydev, line 14: Should be plural

MapleServer2/Packets/CharacterListPacket.cs
```diff
@@ -99,8 +99,8 @@ public static void WriteCharacter(Player player, PacketWriter pWriter)
             pWriter.WriteInt();
             pWriter.WriteShort(player.Levels.Level);
             pWriter.WriteShort();
-            pWriter.WriteInt(player.JobGroupId);
-            pWriter.WriteInt(player.JobId);
+            pWriter.WriteInt((int) player.Job);
```
rhydev, line 102: Can we just change `WriteMode` to `WriteEnum` or make a new method `WriteEnum` that will convert these automatically?

rhydev (changes requested):
Still needs a few changes, honestly there is still more skill stuff to handle and I think some implementations are flawed still, but if there are problems I can just fix them in the future rather than trying to review this over and over.

Sparkymod:
**Reminder:** This a refactoring of how skills is handle, this is yet not finished, as soon the skills is being developed, we will be parsing more skill data, by now nothing is handle besides Id, levels, default levels for each job & each skill for each job. (Spirit is being parse, but needs `AttributeStats` to be implemented.)

**Add:** By request, damage is added to start implementing `Damage`.

MaplePacketLib2/Tools/PacketWriter.cs
```diff
@@ -80,6 +80,11 @@ public PacketWriter WriteMode(Enum value)
             return Write(Convert.ToByte(value));
         }
 
+        public PacketWriter WriteEnum(Enum value)
+        {
+            return Write(Convert.ToInt32(value));
```
rhydev, line 85: This will convert all enums to ints, we need to check what type the enum is and convert to that type.

rhydev (changes requested):

MaplePacketLib2/Tools/PacketWriter.cs
AlanMorel: That or just do `Write(value);`, it accepts basically anything and it should automatically write the size of the Enum, but if not, I agree, this is a useless method if we need to force all enums to be one specific size.

rhydev (approved):

PR #130, opened 2021-01-16
Feat: Multi-threaded GameDataParser

rhydev:
- Added multithreading to GameDataParser
- Added console spinner to indicate it is still running
- Fix ItemParser rarity parsing to default to `1` if rarity not found
- Etc: small fix to GameCommandAction to allow item rarity to be set

### DIFF
--- a/GameDataParser/Program.cs
+++ b/GameDataParser/Program.cs
@@ -1,33 +1,61 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Threading;
+using System.Globalization;
 using System.IO;
 using System.IO.MemoryMappedFiles;
 using System.Xml;
+using System.Collections.Generic;
 using GameDataParser.Crypto;
 using GameDataParser.Crypto.Stream;
 using GameDataParser.Files.Export;
+using GameDataParser.Crypto.Common;
+using GameDataParser.Files;
 
 namespace GameDataParser
 {
     internal static class Program
     {
-        public static ItemMetadataExport Item = new ItemMetadataExport();
-        public static MapMetadataExport MapEntity = new MapMetadataExport();
-        public static SkillMetadataExport Skill = new SkillMetadataExport();
-        public static InsigniaMetadataExport Insignia = new InsigniaMetadataExport();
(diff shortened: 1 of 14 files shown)


GameDataParser/Program.cs
```diff
@@ -1,33 +1,61 @@
...
+            MemoryMappedFile xmlMemFile = MemoryMappedFile.CreateFromFile(VariableDefines.XML_PATH);
+            MemoryMappedFile exportedMemFile = MemoryMappedFile.CreateFromFile(VariableDefines.EXPORTED_PATH);
+
+            // Threads
+            Thread itemThread = new Thread(() => ItemMetadataExport.Export(xmlFiles, xmlMemFile));
+            Thread mapEntityThread = new Thread(() => MapMetadataExport.Export(exportedFiles, exportedMemFile));
+            Thread skillThread = new Thread(() => SkillMetadataExport.Export(xmlFiles, xmlMemFile));
+            Thread insigniaThread = new Thread(() => InsigniaMetadataExport.Export(xmlFiles, xmlMemFile));
+            Thread prestigeThread = new Thread(() => PrestigeMetadataExport.Export(xmlFiles, xmlMemFile));
+            Thread expThread = new Thread(() => ExpMetadataExport.Export(xmlFiles, xmlMemFile));
```
AlanMorel, line 39: You should just create an abstract class that all these exporters extend from so you can define these once and then do 3 for loops, one to make these threads, another to start them, then finally to join them

AlanMorel (changes requested):

angelotadeucci (approved):
Multi-threading! Nice.

GameDataParser/Program.cs
rhydev: Can you give an example of how it would loop I'm not quite understanding, are you saying I should make a list of all the export class objects and then iterate over a loop calling export?

AlanMorel: Nevermind, MapMetadataExport takes in different parameters, which makes changing this not as useful

AlanMorel (approved):

PR #132, opened 2021-01-17
Quest parsing and partial quest handling

angelotadeucci:
### NEEDS REVIEW -- STILL WIP

- Parsing for quest and npc scripts.
- Added Enums for Npc modes.
- Added CompleteQuest and AcceptQuest packets and handling
- Added packet for ContinueChat in NpcTalkPacket

---- Already reviewed ----

-  Added quest parsing
-  Handling of navigator rewards


MapleServer2/Packets/NpcTalkPacket.cs
```diff
@@ -13,10 +13,37 @@ public static Packet Respond(IFieldObject<Npc> npc, int scriptId)
...
-                .WriteInt(9); // DialogType
+                .WriteInt(5); // DialogType
+        }
+
+        private enum NpcType : byte
+        {
+            Dialog = 10,
+        }
+
+        private enum DialogType : int // All this worked with NpcType 10
```
AlanMorel, line 27: Why is this enum private? Seems like it would be broadly useful

MapleServer2/Packets/QuestPacket.cs
```diff
@@ -0,0 +1,70 @@
...
+                }
+            }
+
+            return pWriter;
+        }
+
+        public static Packet CompleteQuest(int questId)
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.QUEST);
+            pWriter.WriteByte((byte) QuestType.CompleteQuest);
```
AlanMorel, line 51: You can use `writeMode` here

AlanMorel (changes requested):

GameDataParser/Parsers/QuestParser.cs
```diff
@@ -0,0 +1,260 @@
...
+                        reader.Skip();
+                    }
+
+                    if (reader.Name == "environment")
+                    {
+                        metadata.Feature = reader["feature"];
+                        metadata.Locale = reader["locale"];
+                    }
+
+                    if (reader.Name == "basic")
```
AlanMorel, line 47: Why not `else if`?

AlanMorel (changes requested):

GameDataParser/Parsers/QuestParser.cs
angelotadeucci: No reason, if you want I can change to `else if`

AlanMorel: Well this code as it stands will keep checking for different values even after it might have already been established earlier, so yeah

AlanMorel (approved):

AlanMorel (approved):

PR #139, opened 2021-01-23
GEM Packet cleanup and ItemPacket improvements.

kvoeten:
GEM works fully again after this, without a hacky copy of the ItemPacket. That said, there's still a bit of a hack in ItemAppearanceInfo, when using the item data from the storage it encodes the wrong colors. More digging into Item Appearance in both the client and the data files is required to fully fix it up. I can see more cases of items not working much like GEM, for which the current hacky fix wouldn't work.


MapleServer2/Packets/GemPacket.cs
```diff
@@ -10,7 +11,18 @@ public static class GemPacket
...
+        }
+
+        public enum GemEquipError : short
+        {
+            InventoryFull,
+            SlotNoLongerValid,
+            ItemNotEligible,
+            ItemExpired,
+            ItemPCCafeOnly,
+            ActivePremiumClubRequired
```
AlanMorel, line 25: Any numerical values for these enums?

MapleServer2/Packets/GemPacket.cs
```diff
@@ -145,6 +48,15 @@ public static Packet UnequipItem(GameSession session, byte index)
             pWriter.WriteInt(session.FieldPlayer.ObjectId);
             pWriter.WriteByte(index);
 
+            return pWriter;
+        }
+        public static Packet EquipError(GemEquipError type)
```
AlanMorel, line 53: Leave empty lines in between methods

MapleServer2/Packets/Helpers/ItemPacketHelper.cs
```diff
@@ -80,6 +93,15 @@ public static PacketWriter WriteItem(this PacketWriter pWriter, Item item)
 
         private static PacketWriter WriteAppearance(this PacketWriter pWriter, Item item)
         {
+            // TODO: make WriteAppearance proper for GEM, this is hacky.
```
AlanMorel, line 96: Just make a `WriteGemAppearance` method, and then do the below check at the calling method, if gem, write gem, if not, write normal appearance. It's not perfect but much less hacky and more readable

AlanMorel (changes requested):

MapleServer2/Packets/GemPacket.cs
kvoeten: They're in a linear array, long as they're ordered numbers are redundant. For readability it's possible to label them 0~5 though. Should I consider numbered Enums as a standard style guide for this project?

MapleServer2/Packets/Helpers/ItemPacketHelper.cs
kvoeten: Personally disagree with this as appearance is one method in the client that matches every item. It's simply not explored yet so the method is partially incorrect. Adding multiple decodes that differ from one another only slightly due to them being for a specific item we don't understand a check for results in redundant code imo.

MapleServer2/Packets/GemPacket.cs
AlanMorel: Nah I was just curious if they had any values at all, I'm cool with them being unlabeled in that case

AlanMorel (approved):

MapleServer2/Packets/ServerListPacket.cs
```diff
@@ -10,18 +10,26 @@ public static class ServerListPacket
...
             foreach (IPEndPoint endpoint in serverIps)
             {
                 pWriter.WriteUnicodeString(endpoint.Address.ToString());
                 pWriter.WriteUShort((ushort) endpoint.Port);
             }
-            pWriter.WriteInt(100); // Const
-            // Looks like length 9, then 1-9 in scrambled order
-            pWriter.Write(09, 00, 01, 00, 04, 00, 07, 00, 02, 00, 05, 00, 08, 00, 03, 00, 06, 00, 09, 00);
+
+            pWriter.WriteInt(0); // 100 in sniff
```
AlanMorel, line 26: If it's 100 in the sniffs, and we still don't know what this does, probably better to keep it 100 even if in this case it doesn't affect anything

MapleServer2/Packets/ServerListPacket.cs
```diff
@@ -10,18 +10,26 @@ public static class ServerListPacket
         public static Packet SetServers(string serverName, ImmutableList<IPEndPoint> serverIps)
         {
             PacketWriter pWriter = PacketWriter.Of(SendOp.SERVER_LIST);
-            pWriter.Write(01, 01, 00, 00, 00);
+
+            pWriter.WriteByte(1); // Required
+            pWriter.WriteInt(0); // Unk.
```
AlanMorel, line 15: Since this is still unknown, maybe just keep it to the old value of 1?

PR #143, opened 2021-01-24
Guild data parsing and other fixes

Zintixx:
Parsing some data for guild contributions and buffs. Submitting this to see if I'm doing it correctly before diving deeper.

Buffs are a bit odd. both personal and guild buffs are in the same table. Personal buffs take meso to use while guild buffs take guild funds. I'm differentiating them by just the id number.

At the moment if you're trying to use a buff, you'll get an error because there's multiple entries of the same buff (due to levels). I'll probably need assistance on how to properly implement guild buffs and their levels. And also calling out the correct buff and level from the parsed data.

Oh, also, I forgot to push out the fieldhandler update from my last PR so stickers can load, so there's that.

### DIFF
--- a/GameDataParser/Files/Export/GuildMetadataExport.cs
+++ b/GameDataParser/Files/Export/GuildMetadataExport.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IO.MemoryMappedFiles;
 using GameDataParser.Crypto.Common;
 using GameDataParser.Parsers;
@@ -10,9 +11,16 @@ namespace GameDataParser.Files.Export
     {
         public static void Export(List<PackFileEntry> files, MemoryMappedFile memFile)
         {
+            if (Hash.CheckHash(VariableDefines.OUTPUT + "ms2-guild-metadata"))
+            {
+                Console.WriteLine("\rSkipping guild metadata!");
+                return;
+            }
+
             // Parse and save some item data from xml file
-            GuildMetadata GuildMetadata = GuildParser.Parse(memFile, files);
-            GuildParser.Write(GuildMetadata);
+            PrestigeMetadata GuildMetadata = PrestigeParser.Parse(memFile, files);
+            PrestigeParser.Write(GuildMetadata);
+            Hash.WriteHash(VariableDefines.OUTPUT + "ms2-guild-metadata");
         }
     }
 }
--- a/GameDataParser/Files/Export/GuildMetadataExport.cs
+++ b/GameDataParser/Files/Export/GuildMetadataExport.cs
@@ -1,0 +1,18 @@
+﻿using System.Collections.Generic;
+using System.IO.MemoryMappedFiles;
+using GameDataParser.Crypto.Common;
+using GameDataParser.Parsers;
+using Maple2Storage.Types.Metadata;
+
+namespace GameDataParser.Files.Export
+{
+    public static class GuildMetadataExport
+    {
+        public static void Export(List<PackFileEntry> files, MemoryMappedFile memFile)
+        {
+            // Parse and save some item data from xml file
+            GuildMetadata GuildMetadata = GuildParser.Parse(memFile, files);
+            GuildParser.Write(GuildMetadata);
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


GameDataParser/Files/Export/GuildMetadataExport.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace GameDataParser.Files.Export
+{
+    public static class GuildMetadataExport
+    {
+        public static void Export(List<PackFileEntry> files, MemoryMappedFile memFile)
+        {
+            // Parse and save some item data from xml file
+            GuildMetadata GuildMetadata = GuildParser.Parse(memFile, files);
+            GuildParser.Write(GuildMetadata);
```
angelotadeucci, line 15: You should update your branch and use the hash feature Rhy added

angelotadeucci (changes requested):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -437,11 +477,29 @@ private void HandleEnterHouse(GameSession session, PacketReader packet)
...
+            if (!session.Player.Wallet.Meso.Modify(-donationAmount))
+            {
+                short NoticeCode = 5121;
+                session.Send(GuildPacket.ErrorNotice(NoticeCode));
+                return;
+            }
+            GuildContribution contribution = GuildMetadataStorage.GetContribution("donation");
+
+            session.Player.GuildContribution += contribution.Value * donateQuantity;
+            int guildFunds = guild.Funds + donateQuantity * 10000;
```
AlanMorel, line 496: You can replace `donateQuantity * 10000` with `donationAmount` here since it's already calculated for you

MapleServer2/Packets/GuildPacket.cs
```diff
@@ -585,7 +574,7 @@ public static Packet DisplayGuildList(List<Guild> guilds)
         {
             PacketWriter pWriter = PacketWriter.Of(SendOp.GUILD);
             pWriter.WriteEnum(GuildPacketMode.DisplayGuildList);
-            pWriter.WriteInt(1); // guild count to display
+            pWriter.WriteInt(guilds.Count); // guild count to display
```
AlanMorel, line 577: You can remove this comment now

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -418,14 +436,36 @@ private void HandleGuildListRequest(GameSession session, PacketReader packet)
...
 
-            session.Send(GuildPacket.ActivateBuff(buffID));
+            GuildBuff buff = GuildMetadataStorage.GetBuff(buffId);
+
+            if (buff.Id > 1000)
+            {
+                if (!session.Player.Wallet.Meso.Modify(-buff.Cost))
+                {
+                    return;
+                };
```
AlanMorel, line 454: You can remove the semi-colon here

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -418,14 +436,36 @@ private void HandleGuildListRequest(GameSession session, PacketReader packet)
...
+            {
+                if (buff.Cost > guild.Funds)
+                {
+                    return;
+                }
+                else
+                {
+                    guild.Funds -= buff.Cost;
+                }
+            }
```
AlanMorel, line 466: The bottom part can look something like this:
```
}
else if (buff.Cost > guild.Funds)
{
    return;
}
               
guild.Funds -= buff.Cost;
```

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
Zintixx: Wouldn't this mean if someone tries to use a buff of an id lower than 1000, they're also paying the cost with guild funds?

angelotadeucci: Yeah if buff id < 1000 they will also pay with the guild funds, aside from that you could remove the else just to make the code cleaner.
Something like this:
```
if (buff.Cost > guild.Funds)
{
    return;
}
 guild.Funds -= buff.Cost;
                
```

angelotadeucci (approved):

PR #149, opened 2021-01-27
GameDataParser architecture overhaul

AlanMorel:


angelotadeucci (approved):

GameDataParser/Files/Hash.cs
```diff
@@ -6,53 +6,40 @@ namespace GameDataParser.Files
...
+            string hashPath = $"{Paths.HASH}/{filename}-hash";
+
+            string newHash = GetHash(filename);
+
+            File.WriteAllText(hashPath, newHash);
+        }
+
+        public static String GetHash(string filename)
+        {
+            using (MD5 md5 = MD5.Create())
```
rhydev, line 35: Since we removed other `using` we should probably remove this as well

GameDataParser/Files/Hash.cs
```diff
@@ -6,53 +6,40 @@ namespace GameDataParser.Files
...
             {
-                using (MD5 md5 = MD5.Create())
+                using (FileStream stream = File.OpenRead(Paths.OUTPUT + "/" + filename))
```
rhydev, line 37: Since we removed other `using` we should probably remove this as well

GameDataParser/Files/MetadataExporter/MetadataExporter.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        private void writeHash()
+        {
+            Hash.WriteHash(this.filename);
+        }
+
+        public void write<Entities>(Entities entities)
+        {
+            using (FileStream writeStream = File.Create(Paths.OUTPUT + "/" + this.filename))
```
rhydev, line 42: Since we removed other `using` we should probably remove this as well

GameDataParser/Parsers/ExpTableParser.cs
```diff
@@ -1,29 +1,27 @@
...
 
 namespace GameDataParser.Parsers
 {
-    class ExpParser
+    class ExpTableParser : Exporter<List<ExpMetadata>>
     {
-        public static List<ExpMetadata> Parse(MemoryMappedFile m2dFile, IEnumerable<PackFileEntry> entries)
+        public ExpTableParser(MetadataResources resources) : base(resources, "exptable") { }
+
+        protected override List<ExpMetadata> parse()
```
rhydev, line 13: The metadata is still named `ExpMetadata`, we should choose to either name everything `ExpTable` or `Exp`

GameDataParser/Files/MetadataExporter/Exporter.cs
```diff
@@ -0,0 +1,20 @@
...
+    public abstract class Exporter<Metadata> : MetadataExporter
+    {
+        protected MetadataResources resources;
+
+        public Exporter(MetadataResources resources, string slug) : base(slug)
+        {
+            this.resources = resources;
+        }
+
+        protected override void serialize()
```
rhydev, line 12: I notice that the new functions are all camel case, it's not necessary for this PR but I think we should decide if we want to change all functions to camel case or not

GameDataParser/Program.cs
```diff
@@ -3,87 +3,51 @@
...
-                Thread itemThread = new Thread(() => ItemMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread mapEntityThread = new Thread(() => MapMetadataExport.Export(exportedFiles, exportedMemFile));
-                Thread skillThread = new Thread(() => SkillMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread insigniaThread = new Thread(() => InsigniaMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread prestigeThread = new Thread(() => PrestigeMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread expThread = new Thread(() => ExpMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread questThread = new Thread(() => QuestMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread scriptThread = new Thread(() => ScriptMetadataExport.Export(xmlFiles, xmlMemFile));
-                Thread guildThread = new Thread(() => GuildMetadataExport.Export(xmlFiles, xmlMemFile));
+            CultureInfo.CurrentCulture = new CultureInfo("en-US");
```
rhydev, line 27: Is there a reason this was moved further down in the code?

GameDataParser/Program.cs
```diff
@@ -3,87 +3,51 @@
...
+            MetadataResources resources = new MetadataResources();
 
-                Spinner spinner = new Spinner();
-                spinner.Start();
+            IEnumerable<MetadataExporter> exports = new List<MetadataExporter>()
```
rhydev, line 30: Perhaps rename to `exporters`

GameDataParser/Program.cs
```diff
@@ -3,87 +3,51 @@
...
-            }
-            catch (Exception ex)
-            {
-                if (ex.ToString().Contains("Xml.m2h") || ex.ToString().Contains("Xml.m2d") || ex.ToString().Contains("Exported.m2h") || ex.ToString().Contains("Exported.m2d"))
-                {
-                    Console.ForegroundColor = ConsoleColor.Red;
-                    Console.WriteLine("Error: Missing pack files.\nPlease copy all pack files (Xml.m2d, Xml.m2h, Exported.m2d, Exported.m2h) from game client into GameDataParser/Resources.");
-                }
-            }
+            Console.WriteLine("\rExporting finished in " + runtime.Minutes + " minutes and " + runtime.Seconds + " seconds");
```
rhydev, line 50: Can we just use a decimal value here? Ex: 9 minutes and 30 seconds => 9.5 minutes.

GameDataParser/Files/Hash.cs
eotetb, line 37: I feel that the use of string interpolation here originally was more readable than changing it to use concatenation. It works either way, but I think the change was unnecessary. It's also inconsistent with the other paths created in this same file now. I recommend changing this one back, or changing all other instances to use concatenation.

rhydev (changes requested):

GameDataParser/Files/MetadataExporter/MetadataExporter.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace GameDataParser.Files
+{
+    public abstract class MetadataExporter
+    {
+        protected string filename;
+
+        public MetadataExporter(string slug)
+        {
+            this.filename = "ms2-" + slug + "-metadata";
```
eotetb, line 13: String concatenation in this file...

GameDataParser/Files/Resources.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public List<PackFileEntry> xmlFiles;
+        public List<PackFileEntry> exportedFiles;
+
+        public MemoryMappedFile xmlMemFile;
+        public MemoryMappedFile exportedMemFile;
+
+        public MetadataResources()
+        {
+            string xmlPath = $"{Paths.INPUT}/Xml.m2d";
```
eotetb, line 18: String interpolation here on the other hand...

GameDataParser/Files/Hash.cs
AlanMorel: The other `using` wasn't in a block but this one is

eotetb (approved):
I left a few minor nitpick comments, but they come down to minor inconsistencies in style more than anything and don't affect the behavior at all. This looks greatly improved and reduces a lot of redundancy that code originally had. The comments I left are worth looking at, but overall I think this is good to merge.

GameDataParser/Files/MetadataExporter/Exporter.cs
AlanMorel: I can change it, coming off of TypeScript

GameDataParser/Parsers/ExpTableParser.cs
AlanMorel: The source already has this inconsistency, inside `ExpMetadataStorage` we have `ms2-exptable-metadata`. This is outside the scope of this PR

GameDataParser/Program.cs
AlanMorel: Not really

GameDataParser/Program.cs
AlanMorel: Why? I'd have to do some math round/floor to support this nicely, and for basically no real upside

GameDataParser/Parsers/ExpTableParser.cs
rhydev: Clearly it's not out of the scope because you renamed ExpParser to ExpTableParser and you removed ExpMetadataExport which is essentially renaming to use ExpTable as well since they are combined now.

AlanMorel: I renamed it here to match the name of the file, since this was the only file that didn't match the name it exported to. Also I removed all MetadataExport files, it had nothing to do with the name

GameDataParser/Files/Hash.cs
AlanMorel: Updated

GameDataParser/Parsers/ExpTableParser.cs
rhydev: So renaming it to match the file for one thing is ok but not for the others, I see.

AlanMorel: Renaming this metadata affects the entire source, not just the parser, which is what this PR is for, how is this that difficult to understand? If I rename the metadata, then it affects the storage, then I have to rename that as well

rhydev: Ok then why didn't you just rename the file to `Exp` then instead of now having more files named `ExpTable`, before the only instance of `exptable` was in the metadata file name, now there are 2 files named that way and 2 files named the other way as opposed to before these changes it was 1 file named that way and 4 files named the other way.

GameDataParser/Program.cs
rhydev: Ok then generally it should be at the top since we're setting a value that is used across the program

GameDataParser/Program.cs
rhydev: That's true, I suppose the current implementation is fine or can simply convert to use `StopWatch.ElapsedMilliseconds` instead of `StopWatch.Elapsed`

GameDataParser/Files/Spinner.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+
+        public void Stop()
+        {
+            active = false;
+            Draw(" ");
+        }
+
+        public TimeSpan getRuntime()
```
rhydev, line 38: Use UpperCamelCase

rhydev (changes requested):
Since we're already changing string concatenation to string interpolation, all new instances of string concatenation should be converted to string interpolation

GameDataParser/Parsers/ExpTableParser.cs
AlanMorel: I changed 25 files this PR, all in GameDataParser only, the title of this PR: " GameDataParser architecture overhaul". I'm not making changes outside of the scope of this PR just because you feel like it. If the issue is that somebody named the file `exptable` but used `Exp` elsewhere, that's not really my problem or the point of this PR. Feel free to tag other developers, they will side with me, this PR has no business touching files that affect the base source

GameDataParser/Program.cs
AlanMorel: Updated

GameDataParser/Files/Spinner.cs
AlanMorel: Updated

rhydev (approved):

GameDataParser/Parsers/ExpTableParser.cs
rhydev: Ok I will make a PR to fix the name + the extra name I have to change now because you changed it to ExpTable.

Sparkymod (approved):

PR #150, opened 2021-01-27
Fix: Refactored parsers

angelotadeucci:
- Renamed ExpTableParser to ExpParser;
- Refactored Insignia, Quest and Exp Parsers to no longer user XmlReader;
- Optimized ScriptParser to ignore events scripts from CN, JP and KR;
- Used dotnet-format;
- New hashes.


GameDataParser/Parsers/ExpParser.cs
```diff
@@ -6,9 +6,9 @@
 
 namespace GameDataParser.Parsers
 {
-    class ExpTableParser : Exporter<List<ExpMetadata>>
+    class ExpParser : Exporter<List<ExpMetadata>>
```
Sparkymod, line 9: Why is this changing again? From `ExpParser` to`ExpTableParser` then again `ExpParser`?

Maple2Storage/Types/Metadata/ScriptMetadata.cs
```diff
@@ -127,29 +127,26 @@ public override string ToString()
     public class Content
     {
         [XmlElement(Order = 1)]
-        public long Script;
```
Sparkymod, line 130: Why is this removed? `Script` must have an Id.

angelotadeucci: You are right, it was a mistake.

angelotadeucci: The script in Content refers to a long that the server doesn't need to know about it.
Example: 
```
    <script randomPick="1" id="10" feature="" locale="">
		<content voiceID="ko/Npc/00000079" **text="$script:0831180407000003$"** />
	</script>
```
The important part is ID for the Element Script not the Content Script.

Sparkymod (approved):

AlanMorel (approved):

PR #151, opened 2021-01-27
Feat: Parser and metadata export for achievements

Andrew-Gan:
### AchieveParser.cs
`protected override List<AchieveMetadata> Parse()`
- utilization of new `Exporter` class to export achievement data as a metadata file
- data read from xml include: achievement id, grade, condition value and reward value

### Program.cs
- registered the new `AchieveParser` into the list of metadata exporters

### AchieveMetadata.cs
- utilization of `SelectNode` and `SelectSingleNode` over `Reader` to avoid iterating over unneeded nodes
- `AchieveMetadata` class includes achievement id (int) and list of `AchieveGradeMetadata` (List<>)
- `AchieveGradeMetadata` class includes achievement grade (or level), condition (goal) and reward (presumably AP earned)

### AchieveMetadataStorage.cs
- Storage class contains functions which will be invoked from Player.cs for updating and sending achievement packets
`public static List<int> GetAchieveIds()`
- return list of IDs of all achievements available in metadata
`AchieveMetadata GetAchieve(int id)`
- return the achievement data associated with the ID
`public static long GetCondition(int id, int grade)`
- return the condition (or goal) needed to reach a certain grade level for a certain achievement
`public static int GetNumGrades(int id)`
- return number of grades available for an achievement


GameDataParser/Parsers/AchieveParser.cs
```diff
@@ -0,0 +1,49 @@
...
+using GameDataParser.Crypto.Common;
+using GameDataParser.Files;
+using Maple2Storage.Types.Metadata;
+
+namespace GameDataParser.Parsers
+{
+    class AchieveParser : Exporter<List<AchieveMetadata>>
+    {
+        public AchieveParser(MetadataResources resources) : base(resources, "achieve") { }
+        protected override List<AchieveMetadata> Parse()
```
AlanMorel, line 12: Can you leave a blank line in between here?

AlanMorel (changes requested):

GameDataParser/Parsers/AchieveParser.cs
Andrew-Gan: you mean between 11 and 12 right? okie

MapleServer2/Data/Static/AchieveMetadataStorage.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace MapleServer2.Data.Static
+{
+    public static class AchieveMetadataStorage
+    {
+        private static readonly Dictionary<int, AchieveMetadata> achieves = new Dictionary<int, AchieveMetadata>();
+
+        static AchieveMetadataStorage()
+        {
+            using FileStream stream = File.OpenRead("Maple2Storage/Resources/ms2-achieve-metadata");
```
angelotadeucci, line 15: Use Paths
Example:
`$"{Paths.RESOURCES}/ms2-guild-metadata"`

angelotadeucci (changes requested):

GameDataParser/Parsers/AchieveParser.cs
AlanMorel: Yeah

angelotadeucci (approved):

AlanMorel (approved):

PR #154, opened 2021-01-28
Fix: Out of bounds teleport on X and Y

angelotadeucci:
Fixed out of bounds to use X and Y coords.
Fixed parsing for map id `02000118` that had two coords for Bounding1.


rhydev (changes requested):
Rather than handling for a specific map id can we have a generic solution?

angelotadeucci:
Changed to check if BoundingBox0 or BoundingBox1 was (0, 0, 0)

GameDataParser/Parsers/MapEntityParser.cs
```diff
@@ -92,20 +92,21 @@ protected override List<MapEntityMetadata> Parse()
                     {
                         XmlNode blockCoord = node.SelectSingleNode("property[@name='Position']");
                         string blockValue = blockCoord?.FirstChild.Attributes["value"].Value;
-
                         CoordS boundingBox = ParseCoord(blockValue ?? "0, 0, 0");
```
rhydev: This code seems duplicated in the if and else if blocks here, can move outside of if and else if blocks.

rhydev (changes requested):

AlanMorel (approved):

MapleServer2/PacketHandlers/Game/UserSyncHandler.cs
```diff
@@ -68,12 +68,26 @@ public override void Handle(GameSession session, PacketReader packet)
 
         private static bool IsOutOfBounds(CoordF coord, CoordS[] boundingBox)
         {
-            CoordS higherBoundingBox = boundingBox[0].Z > boundingBox[1].Z ? boundingBox[0] : boundingBox[1];
-            CoordS lowerBoundingBox = boundingBox[0].Z > boundingBox[1].Z ? boundingBox[1] : boundingBox[0];
-            if (coord.Z > higherBoundingBox.Z || coord.Z < lowerBoundingBox.Z)
+            short HigherBoundZ = boundingBox[0].Z > boundingBox[1].Z ? boundingBox[0].Z : boundingBox[1].Z;
```
AlanMorel, line 71: Why not use math functions here

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/UserSyncHandler.cs
angelotadeucci: Not sure what you mean? You can't do > or < with a Vector3. 
I could only do only two checks instead of six, but I didn't want to trust XML consistency.

rhydev: These variables should be lowerCamelCase also

rhydev: And yeah we can't just > or < on the entire coord object because a Z might be higher in box 0 while having a lower Y, etc.

rhydev (approved):

MapleServer2/PacketHandlers/Game/UserSyncHandler.cs
AlanMorel: But I'm not saying use it on the entire coord object, he's only using it on the Z property here, meaning you should be able to use `Math.Max`:

https://docs.microsoft.com/en-us/dotnet/api/system.math.max?view=net-5.0

AlanMorel (changes requested):

PR #157, opened 2021-01-29
Implemented Equipment and Cosmetic Slots

bwu113:
Database functions for equipping and unequipping regular equipment and cosmetics.
Updated all tablemanager local variables and parameters to comply with camelCase.


Ms2Database/Controllers/EquipmentManager.cs
```diff
@@ -0,0 +1,70 @@
...
+                                             .FirstOrDefault(column => column.EquipmentSlot == slot.ToUpper());
+
+                if (equippedItem == null || equippedItem.EquipmentSlot == "")
+                {
+                    item.EquipmentSlot = slot.ToUpper();
+                }
+                else
+                {
+                    UnequipItem(equippedItem.UniqueId, true);
+                    item.EquipmentSlot = slot.ToUpper();
```
AlanMorel, line 49: Seems like you can move this and line 44 to after the if/else, then maybe invert line 42 to call line 48

AlanMorel (changes requested):

Ms2Database/Controllers/EquipmentManager.cs
bwu113: Gone ahead with the suggestion.

AlanMorel:
Looks good after the merge conflict is fixed

angelotadeucci (approved):
Looks good. Also fix your git history

Issue #88, opened 2020-12-30
NPCs


integralcoder95:
- Scripting
- Proper dialogue (currently can't converse with NPCs, they say one line)
- Quest dialogue

angelotadeucci:
Started with PR #132

Sparkymod:
#169 
#172

Issue #63, opened 2020-12-30
Mobs


integralcoder95:
- Behavior
- Movement AI
- Spawning and Respawning

Sparkymod:
#169 
#171 
#172

angelotadeucci:
#221 
#214

AngeloTadeucci:
#268 
#271

PR #162, opened 2021-02-02
Feat: Parsing all Npc

Sparkymod:
## Npc Parsing
- Add `NpcMetadata` for all entities.
- Add `NpcStats` for NpcMetadata. _The order of this stats are not the same as for `PlayerStats`_
- Remove `MapNpc` and using `NpcMetadata` on MapParsing.

## Npc restruct
- Update `Mob & Npc` class to inherit from `NpcMetadata`.
- Add `AddBoss` for Boss types of Mob. Packet structure is slighly different. _This is added in this PR because is just 1 file being modify._
- Update `FieldManager` to filter Npc from Mobs with `Friendly` property.

Addressing Mobs and Npcs for parsing.
#88 
#63


MapleServer2/Types/Mob.cs
```diff
@@ -1,24 +1,22 @@
...
-        public readonly int Id;
-        public short Rotation; // In degrees * 10
-        public CoordS Speed;
-        public byte Animation;
-        public PlayerStats Stats;
+        //public readonly int Id;
+        public short ZRotation; // In degrees * 10
+        //public CoordS Speed;
+        //public byte Animation;
+        //public PlayerStats Stats;
```
AlanMorel, line 12: Why comment out instead of remove?

AlanMorel (changes requested):

MapleServer2/Types/Mob.cs
Sparkymod: My bad, remove from npc and forgot mob

angelotadeucci (approved):

AlanMorel (approved):

PR #167, opened 2021-02-12
Fix: ItemStats

angelotadeucci:
- Removed rarity option from ItemStats constructor
- Fixed ItemStats for rewards for boxes

### DIFF
--- a/MapleServer2/Tools/GameCommandActions.cs
+++ b/MapleServer2/Tools/GameCommandActions.cs
@@ -93,17 +93,17 @@ namespace MapleServer2.Tools
             }
 
             _ = int.TryParse(config.GetValueOrDefault("rarity", "5"), out int rarity);
+            _ = int.TryParse(config.GetValueOrDefault("amount", "1"), out int amount);
 
-            Item item = new Item(itemId, rarity)
+            Item item = new Item(itemId)
             {
                 CreationTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds(),
                 TransferFlag = TransferFlag.Splitable | TransferFlag.Tradeable,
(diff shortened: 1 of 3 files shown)


MapleServer2/Tools/GameCommandActions.cs
```diff
@@ -93,17 +93,17 @@ private static void ProcessItemCommand(GameSession session, string command)
...
 
-            Item item = new Item(itemId, rarity)
+            Item item = new Item(itemId)
             {
                 CreationTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds(),
                 TransferFlag = TransferFlag.Splitable | TransferFlag.Tradeable,
                 PlayCount = itemId.ToString().StartsWith("35") ? 10 : 0,
+                Rarity = rarity,
+                Amount = amount,
+                Stats = new ItemStats(itemId, rarity)
```
AlanMorel, line 105: Doesn't this already happen by default?

AlanMorel (changes requested):

MapleServer2/Tools/GameCommandActions.cs
angelotadeucci: Unfortunately no, since when Item is instantiated on line 98, it will create ItemStats with the default rarity of the item, some items don't have rarity value in their XML so it's 1. Then after it's intanciated, the Rarity value is changed by the line 103.

angelotadeucci: My old code was using an parameter for rarity on the constructor but I think this way is better.

AlanMorel: Got it

AlanMorel (approved):

Sparkymod (approved):

PR #170, opened 2021-02-13
Feat: Expand inventory & Inventory tab size

angelotadeucci:
### RequestItemInventoryHandler.cs
Added handling for expanding inventory.

### ItemInventoryPacket.cs
Removed hard coded modes, added enum **InventoryMode**;
Added function `Expand()`.

### InventoryController.cs
Added a check on `Inventory.Add()` so packet is not send if item can't be added to inventory;
Added function to `ExpandInventory()`.

### Inventory.cs
Added default and extra size for each inventory tab.

### DIFF
--- a/MapleServer2/Tools/InventoryController.cs
+++ b/MapleServer2/Tools/InventoryController.cs
@@ -182,29 +182,15 @@ namespace MapleServer2.Tools
         {
             Inventory inventory = session.Player.Inventory;
             Wallet wallet = session.Player.Wallet;
-            long meretCost = 390;
+            long meretPrice = 390;
             short expansionAmount = 6;
 
-            if (!wallet.Meret.Modify(-meretCost))
+            if (wallet.RemoveMerets(meretPrice))
             {
-                if (!wallet.GameMeret.Modify(-meretCost))
-                {
-                    if (!wallet.EventMeret.Modify(-meretCost))
-                    {
-                        if (wallet.Meret.Amount + wallet.GameMeret.Amount + wallet.EventMeret.Amount >= meretCost)
-                        {
-                            long rest = wallet.Meret.Amount + wallet.GameMeret.Amount + wallet.EventMeret.Amount - meretCost;
-                            wallet.Meret.SetAmount(rest);
-                            wallet.GameMeret.SetAmount(0);
-                            wallet.EventMeret.SetAmount(0);
-                        }
-                    }
-                }
+                inventory.ExtraSize[tab] += expansionAmount;
+                session.Send(ItemInventoryPacket.LoadTab(tab, inventory.ExtraSize[tab]));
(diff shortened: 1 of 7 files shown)


MapleServer2/Tools/InventoryController.cs
```diff
@@ -169,5 +177,32 @@ public static void Update(GameSession session, long uid, int amount)
             }
             session.Send(ItemInventoryPacket.Update(uid, amount));
         }
+
+        public static void ExpandInventory(GameSession session, InventoryTab tab)
+        {
+            Inventory inventory = session.Player.Inventory;
+            Wallet wallet = session.Player.Wallet;
+
+            if (!wallet.Meret.Modify(-390))
```
AlanMorel, line 186: Can you make 390 a variable?

MapleServer2/Tools/InventoryController.cs
```diff
@@ -169,5 +177,32 @@ public static void Update(GameSession session, long uid, int amount)
...
+                            long rest = wallet.Meret.Amount + wallet.GameMeret.Amount + wallet.EventMeret.Amount - 390;
+                            wallet.Meret.SetAmount(rest);
+                            wallet.GameMeret.SetAmount(0);
+                            wallet.EventMeret.SetAmount(0);
+                        }
+                    }
+                }
+            }
+
+            inventory.ExtraSize[tab] += 6;
```
AlanMorel, line 203: Can you make 6 a variable as well?

AlanMorel (changes requested):

MapleServer2/Tools/InventoryController.cs
```diff
@@ -182,16 +182,18 @@ public static void ExpandInventory(GameSession session, InventoryTab tab)
...
+            if (!wallet.Meret.Modify(-meretCost))
             {
-                if (!wallet.GameMeret.Modify(-390))
+                if (!wallet.GameMeret.Modify(-meretCost))
                 {
-                    if (!wallet.EventMeret.Modify(-390))
+                    if (!wallet.EventMeret.Modify(-meretCost))
                     {
-                        if (wallet.Meret.Amount + wallet.GameMeret.Amount + wallet.EventMeret.Amount >= 390)
+                        if (wallet.Meret.Amount + wallet.GameMeret.Amount + wallet.EventMeret.Amount >= meretCost)
```
AlanMorel, line 194: I wonder if we should have a method inside Wallet that handles removing merets from any of these "currencies", since unless I'm mistaken this won't be the only time we'll need to try deducting from all of them if needed

angelotadeucci: I just tested and Meret Taxi also uses all types of merets.

Sparkymod (approved):

PR #171, opened 2021-02-14
Feat: Adding WIP Damage

Sparkymod:
- Add: a `Priest` character. (it covers almost all situation on skills)
- Add: `SkillCast` class for the skill that is being use in the moment.
- Update: SkillUse, SkillHandler and SkillDamage to use SkillCast.
- Add: `DamageHandler` class container of the damage
- Add: default items for Priest character
- Update: `stats.hp.max = stats.hp.total` for mobs. In order to make calculations, it has to substract/add from `Max`. Then `Total` is used for the hp bar.
- Fix: `SkillJobSub` for those job that has sub skill in `Xml/table/job` but that id isn't an actual existing skill on `Xml/skill`
- Fix: Mobs now disappear from field when they die

### DIFF
--- a/MapleServer2/Packets/SkillDamagePacket.cs
+++ b/MapleServer2/Packets/SkillDamagePacket.cs
@@ -8,28 +8,35 @@ namespace MapleServer2.Packets
 {
     public static class SkillDamagePacket
     {
-        public static Packet ApplyDamage(IFieldObject<Player> player, long skillUid, int someValue, CoordF coords, List<IFieldObject<Mob>> mobs)
+        public static Packet ApplyDamage(IFieldObject<Player> player, long skillSN, int someValue, CoordF coords, List<IFieldObject<Mob>> mobs)
         {
             PacketWriter pWriter = PacketWriter.Of(SendOp.SKILL_DAMAGE);
+            SkillCast skillCast = SkillUsePacket.SkillCastMap[skillSN];
+            DamageHandler damage = DamageHandler.CalculateSkillDamage(skillCast);
+
             pWriter.WriteByte(1);
-            pWriter.WriteLong(skillUid);
+            pWriter.WriteLong(skillSN);
             pWriter.WriteInt(someValue);
             pWriter.WriteInt(player.ObjectId);
             pWriter.WriteInt(player.ObjectId);
-            pWriter.WriteInt(player.Value.ActiveSkillId);
-            pWriter.WriteInt(player.Value.ActiveSkillLevel);
+            pWriter.WriteInt(skillCast.GetSkillId());
+            pWriter.WriteShort(skillCast.GetSkillLevel());
+            // This values appears on some SkillsId, and others like BossSkill, sometimes is 0
+            pWriter.WriteByte(); // Unknown value
+            pWriter.WriteByte(); // Unknown value, also seems to crash if its not a correct value
             pWriter.Write(coords.ToShort());
-            pWriter.Write(CoordS.From(0, 0, 0));
-            /*pWriter.WriteInt(0); // Set as Int because 2 bytes after will set something.
-            pWriter.WriteByte(0); // Define is (59 heal) or (9 damage). Not always the case
-            pWriter.WriteByte(0); // Unknown - count?*/
+            pWriter.Write(CoordS.From(0, 0, 0)); // Position of the image effect of the skillUse, seems to be rotation (0, 0, rotation).
+            // TODO: Check if is a player or mob
             pWriter.WriteByte((byte) mobs.Count); // Mob count
             for (int i = 0; i < mobs.Count; i++)
             {
                 pWriter.WriteInt(mobs[i].ObjectId);
-                pWriter.WriteByte(1); // Unknown
-                pWriter.WriteBool(false); // Crit flag 
-                pWriter.WriteLong(-1 * 1); // TODO: Calculate damage
+                pWriter.WriteByte((byte) damage.GetDamage() > 0 ? 1 : 0); // Unknown
+                pWriter.WriteBool(damage.IsCritical()); // Crit flag 
+                if (damage.GetDamage() != 0)
+                {
+                    pWriter.WriteLong(-1 * (long) damage.GetDamage()); // TODO: Calculate damage
+                }
             }
             return pWriter;
         }
@@ -46,5 +53,6 @@ namespace MapleServer2.Packets
             pWriter.WriteByte(1);
             return pWriter;
         }
+
     }
 }
--- a/MapleServer2/Packets/SkillDamagePacket.cs
+++ b/MapleServer2/Packets/SkillDamagePacket.cs
@@ -8,7 +8,7 @@ namespace MapleServer2.Packets
 {
     public static class SkillDamagePacket
     {
-        public static Packet ApplyDamage(IFieldObject<Player> player, long skillSN, int someValue, CoordF coords, List<IFieldObject<Mob>> mobs)
+        public static Packet ApplyDamage(IFieldObject<Player> player, long skillSN, int unkValue, CoordF coords, List<IFieldObject<Mob>> mobs)
         {
             PacketWriter pWriter = PacketWriter.Of(SendOp.SKILL_DAMAGE);
             SkillCast skillCast = SkillUsePacket.SkillCastMap[skillSN];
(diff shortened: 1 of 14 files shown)


MapleServer2/Packets/SkillDamagePacket.cs
```diff
@@ -46,5 +53,6 @@ public static Packet ApplyHeal(IFieldObject<Player> player, int statusUid, int s
             pWriter.WriteByte(1);
             return pWriter;
         }
+
```
angelotadeucci, line 56: new line

MapleServer2/Packets/StatPacket.cs
```diff
@@ -26,9 +26,12 @@ public static Packet UpdateMobStats(IFieldObject<Mob> mob)
             pWriter.WriteByte(4); // value
             // Stats 
             // Damage should be update through this packet
-            for (int i = 0; i < 3; i++)
+            pWriter.WriteLong(mob.Value.Stats.Hp.Total);
+            pWriter.WriteLong(mob.Value.Stats.Hp.Min);
+            pWriter.WriteLong(mob.Value.Stats.Hp.Max -= (long) skillCast.GetDamage());
+            if (mob.Value.Stats.Hp.Max <= 0)
```
angelotadeucci, line 32: I'm sure we don't want logic inside the packets classes.

MapleServer2/Types/Mob.cs
```diff
@@ -6,12 +6,24 @@ namespace MapleServer2.Types
     public class Mob : NpcMetadata
     {
         public short ZRotation; // In degrees * 10
+        private bool IsDead;
```
angelotadeucci, line 9: Why not use `private bool IsDead {get; set;}` ?

MapleServer2/Types/Player.cs
```diff
@@ -239,7 +237,65 @@ public static Player NewCharacter(byte gender, Job job, string name, SkinColor s
...
+                },
+                IsVIP = false,
+            };
+            player.Equips.Add(ItemSlot.RH, Item.DefaultScepter(player));
+            player.Equips.Add(ItemSlot.LH, Item.DefaultCodex(player));
+            return player;
+        }
+
         public void Warp(MapPlayerSpawn spawn, int mapId)
+
```
angelotadeucci, line 298: new line

MapleServer2/Packets/StatPacket.cs
```diff
@@ -26,9 +26,12 @@ public static Packet UpdateMobStats(IFieldObject<Mob> mob)
             pWriter.WriteByte(4); // value
             // Stats 
             // Damage should be update through this packet
-            for (int i = 0; i < 3; i++)
+            pWriter.WriteLong(mob.Value.Stats.Hp.Total);
+            pWriter.WriteLong(mob.Value.Stats.Hp.Min);
+            pWriter.WriteLong(mob.Value.Stats.Hp.Max -= (long) skillCast.GetDamage());
```
AlanMorel, line 31: can we move this and the below code out the packet method?

MapleServer2/Types/SkillCast.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public short GetSkillLevel()
+        {
+            return SkillLevel;
+        }
+
+        public int GetUnkValue()
+        {
+            return UnkValue;
+        }
```
angelotadeucci, line 34: Also here, why not use `{ get; }` ?

MapleServer2/Types/SkillCast.cs
```diff
@@ -0,0 +1,97 @@
...
+            Random rnd = new Random();
+            double roll = rnd.NextDouble();
+            if (roll > 0.5)
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
```
angelotadeucci, line 60: `return roll > 0.5 ? true : false`

MapleServer2/Types/SkillCast.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            // TODO: Critic Damage calculation
+            if (RollCrit() == true)
+            {
+                return GetDamageRate() + GetDamageRate();
+            }
+            else
+            {
+                return GetDamageRate();
+            }
```
angelotadeucci, line 73: `return RollCrit() ? GetDamageRate() * 2 : GetDamageRate()`

MapleServer2/Types/SkillCast.cs
Sparkymod: sure, forgot that exist.

MapleServer2/Types/Player.cs
```diff
@@ -239,7 +237,65 @@ public static Player NewCharacter(byte gender, Job job, string name, SkinColor s
             };
         }
 
+        public static Player Priest(long accountId, long characterId, string name = "Priest")
+        {
+            Job job = Job.Priest; // Priest
```
AlanMorel, line 242: Not sure we need this comment

MapleServer2/Types/SkillCast.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using MapleServer2.Data.Static;
+
+namespace MapleServer2.Types
+{
+    public class SkillCast
+    {
+        #region Declaration
```
AlanMorel, line 8: Can we remove regions?

AlanMorel (changes requested):

MapleServer2/Types/SkillCast.cs
Sparkymod: sure!, it was a way to understand what i was doing, i can simplify.

angelotadeucci (approved):

MapleServer2/Types/SkillCast.cs
```diff
@@ -50,34 +30,14 @@ public bool RollCrit()
...
             double roll = rnd.NextDouble();
-            if (roll > 0.5)
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
+            return roll > 0.5 ? true : false;
```
AlanMorel, line 33: Don't need the true/false part here

MapleServer2/Packets/StatPacket.cs
```diff
@@ -17,27 +17,23 @@ public static Packet SetStats(IFieldObject<Player> player)
...
             pWriter.WriteByte(4); // value
-            // Stats 
-            // Damage should be update through this packet
-            for (int i = 0; i < 3; i++)
-            {
-                pWriter.WriteLong(mob.Value.Stats.Hp[i]);
-            }
+            pWriter.WriteLong(mob.Value.Stats.Hp.Total);
+            pWriter.WriteLong(mob.Value.Stats.Hp.Min);
+            pWriter.WriteLong(mob.Value.UpdateStats(mob.Value, skillCast.GetDamage()));
```
AlanMorel, line 29: Again, why are we changing mob stats in the middle of the packet? Can this be done outside the packet method? Also, why do we need to pass in `mob.Value` when we already have access to it

MapleServer2/Types/DamageHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public static DamageHandler CalculateSkillDamage(SkillCast skillCast)
+        {
+            double damage;
+            damage = skillCast.GetDamageRate();
+            DamageHandler toReturn = HandleCrit(skillCast, damage);
+            return toReturn;
+        }
+
+
```
AlanMorel, line 45: Blank lines

MapleServer2/Types/Mob.cs
```diff
@@ -1,17 +1,27 @@
...
 using MapleServer2.Data.Static;
 
 namespace MapleServer2.Types
 {
     public class Mob : NpcMetadata
     {
         public short ZRotation; // In degrees * 10
+        public bool IsDead { get; private set; }
+
+        public long UpdateStats(Mob mob, double damage)
```
AlanMorel, line 12: Why's this method above the constructor?

AlanMorel (changes requested):

MapleServer2/Types/Mob.cs
```diff
@@ -25,5 +17,12 @@ public Mob(int id)
             Stats = mob.Stats;
             Friendly = mob.Friendly;
         }
+
+        public void UpdateStats(double damage)
+        {
+            Mob mob = this;
```
AlanMorel, line 23: You don't need this line, you should be able to just access `Stats` and `IsDead` directly here

AlanMorel (changes requested):

MapleServer2/Types/Mob.cs
Sparkymod: sure! yeah, i though i need the same value where is used but is in the same class doesn't need it.

angelotadeucci (approved):

AlanMorel (changes requested):

AlanMorel (approved):

PR #174, opened 2021-02-15
Feat: Dismantle

angelotadeucci:
### ItemParser.cs, ItemMetadata.cs, ItemMetadataStorage.cs
Added `BreakRewards` to Item metadata.

### RequestItemBreakHandler.cs, ItemBreakPacket.cs
Added handling for packets `REQUEST_ITEM_BREAK` and `ITEM_BREAK`.

### FieldEnterHandler.cs [Commit](https://github.com/AlanMorel/MapleServer2/commit/b70630f8d2d481df3b0c1770c53d6b2c77fc9fcb)

Item `20302228` InventoryTab is MISC, but it was adding an invisible item to FRAGMENTS.
I could not find the reason for this issue, so I'm removing the item.


MapleServer2/PacketHandlers/Game/RequestItemBreakHandler.cs
```diff
@@ -0,0 +1,177 @@
...
+namespace MapleServer2.PacketHandlers.Game
+{
+    public class RequestItemBreakHandler : GamePacketHandler
+    {
+        public override RecvOp OpCode => RecvOp.REQUEST_ITEM_BREAK;
+
+        public RequestItemBreakHandler(ILogger<RequestItemBreakHandler> logger) : base(logger) { }
+
+        private Dictionary<short, long> DismantleSlots;
+        private Dictionary<int, int> Rewards;
```
AlanMorel, line 23: While I'm not sure exactly where else this should live, we shouldn't be persisting state in handlers. Feel free to start a convo in Discord or move it to another file (you can also add the helper methods that you are using here)

AlanMorel (changes requested):

MapleServer2/Packets/ItemBreakPacket.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public static Packet ShowRewards(Dictionary<int, int> rewards)
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.ITEM_BREAK);
+            pWriter.WriteEnum(ItemBreakMode.ShowRewards);
+            pWriter.WriteByte(1); // unknown
+            pWriter.WriteInt(rewards.Count);
+            foreach (KeyValuePair<int, int> item in rewards)
+            {
```
Sparkymod, line 59: did you get the case where there's no items but still send this packet? breaking just use items ? or can handle filter to not break "potions"

angelotadeucci: Do you mean if I send this packet and rewards is count = 0?

Sparkymod: yes

AlanMorel (approved):

Sparkymod (approved):

PR #176, opened 2021-02-15
Fix: ItemStats

angelotadeucci:
- Renamed attributes.
- Skip parsing if item rarity has 0 OptionsNumPick.
- Using Linq instead of for loops.


Sparkymod (approved):

MapleServer2/Types/ItemStats.cs
```diff
@@ -109,28 +109,28 @@ public ItemStats(int itemId, int grade)
                 }
             }
 
-            if (ItemStatsMetadataStorage.GetRandomStat(itemId, out List<ItemOptions> randomList))
+            if (ItemStatsMetadataStorage.GetRandomBonus(itemId, out List<ItemOptions> randomBonusList))
             {
-                foreach (ItemOptions item in randomList)
+                foreach (ItemOptions options in randomBonusList)
```
sgessa, line 114: Instead of iterating over all the list you could use:

```suggestion
randomBonusList.Find(options => options.Rarity == rarity && options.Slots > 0));
```

We dont want multiple options, just the one with the same rarity and at least 1 slot.
Return immediately if no options is found.

sgessa: Same principle apply to `basicList`

MapleServer2/Types/ItemStats.cs
```diff
@@ -109,28 +109,28 @@ public ItemStats(int itemId, int grade)
...
+                    if (options.Slots == 0)
                     {
                         continue;
                     }
-                    List<int> indexes = GetRandomOptions(item.OptionNumPick, item.Stats.Count);
+                    List<int> indexes = GetRandomOptions(options.Slots, options.Stats.Count);
                     foreach (int i in indexes)
                     {
-                        if (item.Stats[i].Value != 0)
+                        if (options.Stats[i].Value != 0)
```
sgessa, line 127: You could probably write a method to reuse this code, you're doing the same thing for basic attributes.

sgessa (changes requested):

sgessa (approved):

PR #181, opened 2021-02-18
Fix: Dismantle

angelotadeucci:
- New function `Consume` in `InventoryController.cs`.
- Fixed removing the correct amount of items from inventory.
- Fixed giving the correct amount of rewards.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/RequestItemBreakHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/RequestItemBreakHandler.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using Maple2Storage.Types;
 using Maple2Storage.Types.Metadata;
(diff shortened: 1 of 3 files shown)


MapleServer2/PacketHandlers/Game/RequestItemBreakHandler.cs
```diff
@@ -67,9 +68,20 @@ private static void HandleAdd(GameSession session, PacketReader packet)
         private static void HandleRemove(GameSession session, PacketReader packet)
         {
             long uid = packet.ReadLong();
-            short key = session.Player.DismantleSlots.FirstOrDefault(x => x.Value == uid).Key;
+            Tuple<long, int>[] dismantleSlots = session.Player.DismantleSlots;
+            for (int i = 0; i < dismantleSlots.Length; i++)
```
sgessa, line 72: You can probably just find the index having a null slot using Linq

sgessa (changes requested):

MapleServer2/PacketHandlers/Game/RequestItemBreakHandler.cs
angelotadeucci: Not sure what you mean by null slot.

angelotadeucci: I tried but array doesn't have Find?

sgessa (approved):

PR #191, opened 2021-02-24
Chat Sticker expiration parsing

Zintixx:
- Revised parsing for function parameters to consider more parameters than just id
- Included other function parameter data parsed. 
- Moved chat stickers to it's own class

Note this is not all the function parameters have been parsed yet. A lot of different items use the function parameter field as a catch all for any kind of parameters (as you can see in boxes, mini game events (OpenMassive), and stickers have all different ways it's being parsed). I wasn't sure how exactly to approach this so I went by adding on to what we already have created. Would love some more guidance on this.


angelotadeucci (approved):

GameDataParser/Parsers/ItemParser.cs
```diff
@@ -211,8 +211,32 @@ protected override List<ItemMetadata> Parse()
...
+
+                    int durationSec = 0;
+
+                    if (functionParameters.Attributes["durationSec"] != null)
+                    {
+                        durationSec = int.Parse(functionParameters.Attributes["durationSec"].Value);
+                    }
+                    metadata.FunctionDuration = durationSec;
+                }
+                if (contentType == "OpenMassive")
```
AlanMorel, line 228: Could be `else if`

AlanMorel (changes requested):

angelotadeucci (approved):

PR #212, opened 2021-03-12
Spawn block offset and rotation fix

Zintixx:
Fix for moving fields. It should now behave like it does in the game:

- Rotation added to face the correct way
- If the portal was a cube , you spawn one block away from the cube rather than on it. The spawn block depends on the rotation.
- I've also removed a null skip on "TargetFieldSN" in MapEntityParser. This is important for same-map teleporting.


MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
```diff
@@ -46,8 +46,31 @@ public override void Handle(GameSession session, PacketReader packet)
 
                 // TODO: There needs to be a more centralized way to set coordinates...
                 session.Player.MapId = srcPortal.Target;
+                session.Player.Rotation = dstPortal.Rotation.ToFloat();
                 session.Player.Coord = dstPortal.Coord.ToFloat();
-                session.Player.SafeBlock = dstPortal.Coord.ToFloat();
+
+                if (dstPortal.Name == "Portal_cube") // spawn on the next block if portal is a cube
+                {
+                    int offset = 150;
```
AlanMorel, line 54: I think 150 is already a constant somewhere

AlanMorel (changes requested):

AlanMorel (approved):

angelotadeucci (approved):

PR #217, opened 2021-03-20
Small fixes

angelotadeucci:
- **InteractActorPacket**
    - Only write extra int if `InteractActorType` is **Extractor** or **Gathering**;
- **FieldObjectPacket**
    - Small fix on **LoadPlayer** packet;
    - Added packet to remove player from map;
- **FieldManager**
    - Added `FieldObjectPacket.RemovePlayer()`;
- **InventoryController**
   - `Consume()` don't remove item if amount changed is greater than amount owned;
   - Removed `Update()` function;
   - Updated `MasteryHandler` and `RequestItemUseMultipleHandler` to use `Consume()`.
- **Dotnet-format**

### DIFF
--- a/MapleServer2/PacketHandlers/Game/RequestItemUseMultipleHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/RequestItemUseMultipleHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
 using MapleServer2.PacketHandlers.Game.Helpers;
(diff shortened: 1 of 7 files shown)


AlanMorel (approved):

MapleServer2/PacketHandlers/Game/RequestItemUseMultipleHandler.cs
```diff
@@ -39,43 +40,27 @@ public override void Handle(GameSession session, PacketReader packet)
...
-
+                Item item = kvp.Value;
                 // Do nothing if box has no data stored
                 if (item.Content.Count <= 0)
                 {
                     break;
                 }
 
-                for (int i = opened; i < amount; i++)
+                for (int i = 0; i < amount; i++)
```
rhydev, line 54: This is incorrect, without keeping track of number of opened boxes it'll keep resetting the amount of boxes to open when it goes to the next stack of boxes.

MapleServer2/Packets/FieldObjectPacket.cs
```diff
@@ -21,19 +22,19 @@ public static Packet LoadPlayer(IFieldObject<Player> fieldPlayer)
...
             pWriter.WriteByte();
             pWriter.Write(fieldPlayer.Coord);
             pWriter.WriteShort(player.Levels.Level);
-            pWriter.WriteEnum(player.Job); // short
-            pWriter.WriteEnum(player.JobCode);
-            pWriter.WriteInt();
-            pWriter.WriteInt();
-            pWriter.WriteInt();
+            pWriter.WriteShort((short) player.Job);
+            pWriter.WriteShort((short) player.JobCode);
```
rhydev, line 34: This should be using `WriteEnum`, not sure why changed.

rhydev (changes requested):

Sparkymod (commented):
should be good, as soon the requested changes are resolve

MapleServer2/Packets/FieldObjectPacket.cs
angelotadeucci: Because it's needed to write a short and most other places Job is used as int.

rhydev (approved):

PR #221, opened 2021-03-23
Feat: Mob respawn

quotopian:
## Additions:
- Parse spawn point IDs.
- Mob respawns according to its own cooldown timer.
- **MobSpawn.cs**
   - Moves mob spawns to the FieldManager.
   - Keeps track of its own mobs.
   - Contains a spawn point ID.

## Removals
- **SpawnGenerator.cs**
   - Moved methods into MobSpawn.


angelotadeucci (approved):

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -396,5 +386,39 @@ private Task StartHealingSpot(GameSession session, IFieldObject<Player> player)
...
+
+            foreach (NpcMetadata mob in mobSpawn.Value.SpawnMobs)
+            {
+                int spawnCount = mob.NpcMetadataBasic.GroupSpawnCount;  // Spawn count changes due to field effect (?)
+                if (spawnCount > mobSpawn.Value.MaxPopulation)
+                {
+                    break;
+                }
+
+                for (int i = 0; i < spawnCount && i < mobSpawn.Value.MaxPopulation; i++)
```
AlanMorel, line 402: Because of the check on line 397, why are both of these i checks needed? `spawnCount` should always be less meaning the second check seem useless to me?

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -396,5 +386,39 @@ private Task StartHealingSpot(GameSession session, IFieldObject<Player> player)
...
+                    AddMob(fieldMob);
+                }
+            }
+        }
+
+        private Task StartSpawnTimer(IFieldObject<MobSpawn> mobSpawn)
+        {
+            return Task.Run(async () =>
+            {
+                await Task.Delay(mobSpawn.Value.SpawnData.SpawnTime * 1000);
```
AlanMorel, line 415: Might be wrong but if you move this to before the return line, it only needs to do this multiplication one time

AlanMorel (changes requested):

MapleServer2/Servers/Game/FieldManager.cs
quotopian: I think it still does the multiplication one time since the function only runs once.

If you mean if the compiler would make it a constant, I think it would still recognize it as a constant in this case too, but I'm not sure as well. I can change it.

PR #222, opened 2021-03-23
Feat + Fix: Trophy tracking and reward handling

Andrew-Gan:
Set up the code structure for trophy tracking and reward handling.
Replaced all instances of 'Achieve' with 'Trophy' to follow in-game naming.

### RewardType.cs
- representing reward types using byte enum for smaller metadata size
- developers working on other parts of code that handles reward type MAY use the enum for reading from metadata file

### TrophyParser.cs
- include information related to rewards

### MasteryMetadataStorage.cs
- new function to convert XP into corresponding trophy level

### Trophy.cs
- fixed trophy level offset error in
```C
void AddCounter(GameSession, long)
```
- handle rewards automatically when player trophy levels up


MapleServer2/Data/Static/MasteryMetadataStorage.cs
```diff
@@ -29,5 +30,16 @@ public static MasteryMetadata GetMastery(int type)
...
+        public static int GetGradeFromXP(MasteryType type, long xp)
+        {
+            int grade = 0;
+            List<MasteryGrade> masteries = GetMastery((int) (type)).Grades;
+            while ((grade < masteries.Count) && (xp >= masteries[grade].Value))
+            {
+                grade++;
+            }
+            return grade;
+        }
```
angelotadeucci, line 43: Linq is good. Linq is your friend.

```suggestion
        public static int GetGradeFromXP(MasteryType type, long xp)
        {
            return GetMastery((int) type).Grades.Count(x => x.Value < xp);
        }
```

MapleServer2/Data/Static/TrophyMetadataStorage.cs
```diff
@@ -0,0 +1,47 @@
...
+            return map.GetValueOrDefault(id);
+        }
+
+        public static TrophyGradeMetadata GetGrade(int id, int grade)
+        {
+            if ((grade < 1) || (grade > GetNumGrades(id)))
+            {
+                return null;
+            }
+            return map.GetValueOrDefault(id).Grades[grade - 1];
```
angelotadeucci, line 39:
```suggestion
            return map.GetValueOrDefault(id).Grades.FirstOrDefault(x => x.Grade == grade);
```

MapleServer2/Packets/TrophyPacket.cs
```diff
@@ -0,0 +1,77 @@
...
+                pWriter.WriteInt(1);            // unknown 'SS' ?
+                WriteIndividualTrophy(pWriter, trophy);
+            }
+
+            return pWriter;
+        }
+
+        public static Packet WriteUpdate(Trophy trophy)
+        {
+            PacketWriter pWriter = PacketWriter.Of(SendOp.ACHIEVE);
```
angelotadeucci, line 51: Maybe rename the SendOp and RecvOp too, but I'm not sure about that.

MapleServer2/Types/Trophy.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    Condition = TrophyMetadataStorage.GetGrade(Id, NextGrade).Condition;
+                }
+                // level up and completed
+                else
+                {
+                    IsDone = true;
+                    NextGrade--;
+                    string[] cats = TrophyMetadataStorage.GetMetadata(Id).Categories;
+                    foreach (string cat in cats)
```
angelotadeucci, line 61: Maybe rename cats to `categories` and cat to `categorie`? Because it doesn't look better simplified like that

MapleServer2/Types/Trophy.cs
```diff
@@ -0,0 +1,122 @@
...
+                case RewardType.item:
+                    break;
+                case RewardType.shop_weapon:
+                    break;
+                case RewardType.dynamicaction:
+                    break;
+                case RewardType.etc:
+                    break;
+                case RewardType.beauty_skin:
+                    break;
```
angelotadeucci, line 116:
```suggestion
                case RewardType.Unknown:
                case RewardType.itemcoloring:
                case RewardType.shop_ride:
                case RewardType.title:
                case RewardType.beauty_hair:
                case RewardType.skillPoint:
                case RewardType.beauty_makeup:
                case RewardType.shop_build:
                case RewardType.item:
                case RewardType.shop_weapon:
                case RewardType.dynamicaction:
                case RewardType.etc:
                case RewardType.beauty_skin:
                    break;
                case RewardType.statPoint:
                    session.Player.StatPointDistribution.AddTotalStatPoints(grade.RewardValue, OtherStatsIndex.Trophy);
                    session.Send(StatPointPacket.WriteTotalStatPoints(session.Player));
                    break;
                default:
                    break;
```

angelotadeucci (changes requested):

MapleServer2/Packets/TrophyPacket.cs
Andrew-Gan: yeah coz i think those are the same as the ones in MapleShark2 so I might keep them that way for now

MapleServer2/Types/Trophy.cs
Andrew-Gan: it will ultimately require breaks at every line so are you suggesting we should keep the code clean while I wait for other parts to be implemented?

MapleServer2/Data/Static/MasteryMetadataStorage.cs
Andrew-Gan: ok hail C#

MapleServer2/Types/Trophy.cs
angelotadeucci: I think it's better to keep the code clean since we don't know when everything else will be implemented!

MapleServer2/Packets/TrophyPacket.cs
angelotadeucci: Yes but it's like the discussion between ingame and xml names. We should use the ingame name always right? So between ingame and mapleshark I would prefer to use ingame terminology

c:\temp\maple-server-internal.log
```diff
@@ -0,0 +1,10 @@
+2021-03-23 17:59:11.8452 Info Auto loading assembly file: /home/andrew-gan/Documents/git/MapleServer2/MapleServer2/bin/Debug/net5.0/NLog.Extensions.Logging.dll
```
angelotadeucci, line 1: What is this

angelotadeucci (changes requested):

MapleServer2/Packets/TrophyPacket.cs
Andrew-Gan: i see that other opcodes are named that way as well (ex. STAT for StatPacket.cs). Modified to follow convention.

c:\temp\maple-server-internal.log
Andrew-Gan: running the code typically outputs a log file in windows but the path is invalid on linux so it was created here instead.
Removed :p

angelotadeucci (approved):

GameDataParser/Parsers/TrophyParser.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace GameDataParser.Parsers
+{
+    class TrophyParser : Exporter<List<TrophyMetadata>>
+    {
+        public TrophyParser(MetadataResources resources) : base(resources, "trophy") { }
+
+        protected override List<TrophyMetadata> Parse()
+        {
+            HashSet<string> test = new HashSet<string>();
```
AlanMorel, line 17: ?

MapleServer2/Types/Player.cs
```diff
@@ -118,14 +118,29 @@ public class Player
...
+        private Task OnlineDurationThread;
+        private TimeInfo Timestamps;
         public Dictionary<int, PlayerStat> GatheringCount = new Dictionary<int, PlayerStat>();
 
+        class TimeInfo
+        {
+            public long charCreation = -1;
+            public int onlineDuration = 0;
+            public long lastOnline = -1;
+            public TimeInfo(long a = -1, int b = 0, long c = -1)
```
AlanMorel, line 130: Line before constructor

MapleServer2/Types/Player.cs
```diff
@@ -118,14 +118,29 @@ public class Player
...
-
+        private Task OnlineDurationThread;
+        private TimeInfo Timestamps;
         public Dictionary<int, PlayerStat> GatheringCount = new Dictionary<int, PlayerStat>();
 
+        class TimeInfo
+        {
+            public long charCreation = -1;
+            public int onlineDuration = 0;
+            public long lastOnline = -1;
```
AlanMorel, line 129: You don't need these defaults here if you're already setting them in the constructor parameters

AlanMorel (changes requested):

GameDataParser/Parsers/TrophyParser.cs
Andrew-Gan: Leftover code from debug removed

MapleServer2/Types/Player.cs
```diff
@@ -124,14 +124,15 @@ public class Player
...
         {
-            public long charCreation = -1;
-            public int onlineDuration = 0;
-            public long lastOnline = -1;
-            public TimeInfo(long a = -1, int b = 0, long c = -1)
+            public long CharCreation;
+            public long OnlineDuration;
+            public long LastOnline;
+
+            public TimeInfo(long a = -1, long b = 0, long c = -1)
```
AlanMorel, line 131: can a, b, c be given descriptive names?

AlanMorel (changes requested):

PR #218, opened 2021-03-20
Feat: Buddy List and Music Compose

Zintixx:
## Buddy List
Implemented Buddy List. This also includes partial Block functionality. 
Buddy system works by a shared ID system. Both players are connected by the ID in order to get each other's information on each Buddy List.
There's a few things in the current implementation that are temporary (as indicated as comments) until a database is set up.

## Music Compose
Adjusted the WriteItem packet to work with UGC Scores.
Parsing "isCustomNote" to find the items that are UGCScores.
Music composing should now work as intended. 
Listening to music for _other_ players in the map still does not work. I left a comment as to what needs to be fixed.

### Etc
Moved all of the function Item data to it's own class. This should make it a bit cleaner.

PR #223, opened 2021-03-26
Fix #222

angelotadeucci:


angelotadeucci:
Master will get fixed with PR #218

PR #224, opened 2021-03-27
Feat: Meret Market & Gacha Capsules

Zintixx:
## Meret Market
Implemented Meret Market - only the home page and premium section (no UGC/Design shop).
The home page heavily relies on banner images from a web server however there were some default in-game 'banners' that are able to be used.
Multi item buying (through the cart) is added.

## Gachapon/Capsules
Capsules have been implemented. Ducky Outfit Capsules tend to have 'SmartGender' which means anything a player gets will be their gender. In order to achieve this, I have a do while loop to keep looping until it randomly picks a gender of the character.
Capsule shops have _not_ been implemented. This requires knowing the exact Shop category string in order to correctly have it display in the shop.  I'm not sure how we could achieve this.
Outfits/Items obtained from capsules cannot be broken down to coins has not been implemented.

### Various other fixes
Added a shop type enum. This will tell the client what shop "skin" to use.
Added PercussionId to the `InstrumentPacket PlayScore` packet. This allows for percussion instruments to properly play their sound.
Adjusted Super World Chat to use the item's ID rather than the function ID. This fix was needed for item copies of the same item but different IDs.
EDIT: Also added GameEvent. The client must have some coded in game events that recognizes the string sent by server. At the moment, this is really down the line that should be implemented. However, the only one I added at the moment was StringBoard. This is just the marquee seen in game.


MapleServer2/PacketHandlers/Game/MeretMarketHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            Item item = new Item(marketItem.ItemId)
+            {
+                Amount = marketItem.Quantity + marketItem.BonusQuantity,
+                Rarity = marketItem.Rarity
+            };
+            if (marketItem.Duration != 0)
+            {
+                item.ExpiryTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds() + Environment.TickCount + (marketItem.Duration * 86400);
```
AlanMorel, line 133: Can the 86400 be broken down into further smaller multiplication so its more obvious what it represents?

MapleServer2/PacketHandlers/Game/RequestItemUseHandler.cs
```diff
@@ -192,5 +197,77 @@ private static void HandleHongBao(GameSession session, Item item)
...
+            }
+        }
+
+        private static GachaContent HandleSmartGender(GachaMetadata gacha, byte playerGender)
+        {
+            Random random = new Random();
+            int index = random.Next(gacha.Contents.Count);
+
+            GachaContent contents = gacha.Contents[index];
+            if (contents.SmartGender)
```
AlanMorel, line 249: You can invert and immediately return contents

AlanMorel (changes requested):

AlanMorel (approved):

Sparkymod (approved):

angelotadeucci (approved):

PR #230, opened 2021-04-05
Fix: Attack on npcStats should be Accuracy.

gugarosa:
This PR proposes just a minor change that I almost overlooked when understanding the parsers. An `atp` tag indicates `Accuracy` instead of `Attack`, in the same way it is proposed by `ItemOptionRangeParser`.

Additionally, I have checked some of the values in the parsed output and they indeed correspond to the accuracy attribute, e.g., [slime](https://maplestory2.fandom.com/wiki/Slime) has 80 acc instead of 80 atk.


angelotadeucci (approved):

AlanMorel (approved):

Zintixx:
yep, looks correct:
`<key id="s_word_stat_atp" value="Accuracy" />`

Zintixx (approved):
yep, looks correct:
<key id="s_word_stat_atp" value="Accuracy" />

Sparkymod (approved):

PR #246, opened 2021-04-21
Feat: Ad Balloons

Zintixx:
Implements ad balloons. 
~~Created a separate InteractObject class for this, as it has a lot of different variables that normal InteractObjects do not have.~~
EDIT: Ad Balloon object class is now combined with the standard InteractObject class to ensure InteractObjectHandler is cleaner.


MapleServer2/PacketHandlers/Game/InteractObjectHandler.cs
```diff
@@ -56,6 +56,14 @@ private static void HandleStart(GameSession session, PacketReader packet)
         private static void HandleUse(GameSession session, PacketReader packet)
         {
             string uuid = packet.ReadMapleString();
+
+            if(uuid.StartsWith("BillBoard")) // Possible temp solution?
```
AlanMorel, line 60: Space after the `if`

MapleServer2/PacketHandlers/Game/InteractObjectHandler.cs
```diff
@@ -56,6 +56,14 @@ private static void HandleStart(GameSession session, PacketReader packet)
...
+            {
+                IFieldObject<AdBalloon> balloon = session.FieldManager.State.Balloons[uuid];
+                session.Send(PlayerHostPacket.AdBalloonWindow(balloon));
+                return;
+            }
+
```
Zintixx, line 66: Highlighting this portion for conversation:
At the moment, we're checking if an interact object exists in the field by MapEntityStorage. Of course, this wont work for balloons because they're user-placed objects. 
My currently solution right now to differentiate the two is by the name. _IDEALLY_ I think we should go by InteractObjectType.
Instead of searching MapEntityStorage for the InteractObject, should we instead search the Field?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/InteractObjectHandler.cs
AngeloTadeucci: The best would be to rework _**InteractObjectHandler**_ to use `FieldState` instead of **_MapEntityStorage_**. Then with the enum _InteractObjectType_ it should also handle future objects that are added through the instance like instance portals (not sure the name, portals that help you navigate quests)

Maple2Storage/Types/Metadata/ItemMetadata.cs
```diff
@@ -240,6 +240,16 @@ public class ItemFunction
...
+        [XmlElement(Order = 10)]
+        public string Model;
+        [XmlElement(Order = 11)]
+        public string Asset = "";
+        [XmlElement(Order = 12)]
+        public string NormalState;
+        [XmlElement(Order = 13)]
+        public string Reactable;
+        [XmlElement(Order = 14)]
+        public float Scale = 1;
```
AngeloTadeucci, line 252: If these attributes are **only** used by ad balloons maybe do a separated class?

MapleServer2/PacketHandlers/Game/RequestItemUseHandler.cs
```diff
@@ -346,5 +349,22 @@ public static void HandleCallAirTaxi(GameSession session, PacketReader packet, I
                 session.Player.Warp(spawn, fieldID);
             }
         }
+
+        public static void HandleInstallBillBoard(GameSession session, PacketReader packet, Item item)
+        {
+            string[] parameters = packet.ReadUnicodeString().Split("'");
+            string title = parameters[0];
+            string description = parameters[1];
+            bool publicHouse = parameters[2].Equals("1") ? true : false;
```
AngeloTadeucci:
```suggestion
            bool publicHouse = parameters[2].Equals("1");
```

MapleServer2/PacketHandlers/Game/InteractObjectHandler.cs
AngeloTadeucci: The best would be to rework InteractObjectHandler to use FieldState instead of MapEntityStorage. Then with the enum InteractObjectType it should also handle future objects that are added through the instance like instance portals (not sure the name, portals that help you navigate quests)

Zintixx: Yeah that's what I was thinking. And have a switch case for the InteractObjectType.
Would this mean AdBalloons need to be combined with "normal" InteractObject?

AngeloTadeucci: I think it's the best way unless there is a reason to AdBallons be its own `type`. If InteractObject and AdBalloon have way too different implementations then they should be separate, which I don't think it's the case

AngeloTadeucci (approved):

AlanMorel (approved):

Sparkymod (approved):

PR #247, opened 2021-04-23
WIP: Database

AngeloTadeucci:
### Initial implementation for database using MySQL
- Implemented basic account and character CRUD;
- Implemented most tables related to `Player`.

You can find all tables and columns in the diagram [here.](https://user-images.githubusercontent.com/15664821/115944795-6e6ed280-a48e-11eb-9630-cd357a93e817.png)

### Code changes
- Added support for **.env** file;
- Removed `AccountStorage`;
- Renamed attributes so they are more explicit;
- Moved `Equips`, `Cosmetics` and `Badges` from **Player** to **Inventory**;

### Looking for feedback since it's my first time using **EntityFramework** and any type of ORM. Thank you :)

Quick guide to make it work:

1. Install MariaDB or MySQL driver.
2. Create `.env` file using the `.env.example` as example and set correct user name and password for your database.
3. Run `MapleServer2`.
4. ???
5. ~~complain that items aren't being saved~~


MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
```diff
@@ -98,100 +110,101 @@ public void HandleCreate(LoginSession session, PacketReader packet)
 
                         equips.Add(ItemSlot.HR, new Item(Convert.ToInt32(id))
                         {
-                            CreationTime = 1565575851,
+                            CreationTime = DateTimeOffset.Now.ToUnixTimeSeconds(),
```
AlanMorel, line 113: This should just be the default value inside `Item`

MapleServer2/Types/Currency.cs
```diff
@@ -39,6 +42,7 @@ public void SetAmount(long input)
 
         private void UpdateWallet()
         {
+            DatabaseManager.Update(Player.Wallet);
```
AlanMorel, line 45: This feels dicey, it means everytime meso is changed in any way (or any currency), it will make a database call when instead it should be enough to save upon a channel/map change or logging out, and keep the new wallet state in memory

MapleServer2/Types/Inventory.cs
```diff
@@ -52,6 +61,34 @@ public Inventory(IEnumerable<Item> loadItems) : this()
...
+            Badges = new List<Item>();
+            Id = inventory.Id;
+            ExtraSize = inventory.ExtraSize;
+            foreach (Item item in inventory.DB_Items)
+            {
+                if (item.IsEquipped)
+                {
+                    Equips.Add(item.ItemSlot, item);
+                }
+                else if (item.IsEquipped && item.InventoryTab == InventoryTab.Outfit)
```
AlanMorel, line 77: Don't need to keep checking if the item is equipped on 77 and 81 because if it is, it will be true on line 73

MapleServer2/Types/Item.cs
```diff
@@ -154,7 +151,7 @@ public static Item Ear()
         {
             return new Item(10500001)
             {
-                Uid = 2754959794416496488,
+                Uid = GuidGenerator.Long(),
```
AlanMorel, line 154: This might already be the default behavior

MapleServer2/Types/Levels.cs
```diff
@@ -100,6 +109,7 @@ public void GainExp(int amount)
 
             Exp = newExp;
             Player.Session.Send(ExperiencePacket.ExpUp(amount, newExp, RestExp));
+            DatabaseManager.Update(this);
```
AlanMorel, line 112: We can talk in Discord but this is dicey for the same reason as before. Database calls should not be made on things that will happen this often

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/QuitHandler.cs
```diff
@@ -12,25 +13,47 @@ namespace MapleServer2.PacketHandlers.Game
...
-                session.FieldManager.RemovePlayer(session, session.FieldPlayer);
-                AuthData authData = AuthStorage.GetData(session.Player.AccountId);
-
-                session.SendFinal(MigrationPacket.GameToLogin(LoginEndpoint, authData));
+                case QuitMode.ChangeCharacter:
+                    HandleChangeCharacter(session);
+                    break;
+                case QuitMode.Quit:
+                    HandleQuit(session);
+                    break;
```
AlanMorel, line 40: Can this get a default logger?

AlanMorel (approved):

Sparkymod (approved):
👍

MapleServer2/PacketHandlers/Game/QuitHandler.cs
AngeloTadeucci: Done

PR #251, opened 2021-04-30
Add configuration

darkvergus:
Added basic configuration still missing how to get the ip address to work so i couldn't get it to work atm though its in the config

to create a new section for the configuration such as xp rates all its required is to create a file like the example i used for ServerConfig and extend ConfigBase

image

After that just add the config to the configuration file like so

image

and you're ready to use it.

ATM it just supports floats, ints, bools and strings but feel free to add more types if you need.


AngeloTadeucci (approved):

MapleServer2/Config/ConfigManager.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        public static Configuration LoadFromIni(Stream iniStream)
+        {
+            using (StreamReader iniReader = new StreamReader(iniStream))
+            {
+                FileIniDataParser parser = new FileIniDataParser();
+                IniData configdata = parser.ReadData(iniReader);
+
+                Configuration conf = new Configuration();
```
AlanMorel, line 66: Looks like lines 43-57 and 66-78 are identical, could probably make a shared helper method and call that instead

AlanMorel (changes requested):

rhydev:
If we're going to add config we should use config files.

MapleServer2/Config/ConfigHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            this.commentDelimiter = commentDelimiter;
+            this.file = file;
+        }
+
+        public void Init(string value)
+        {
+            file = null;
+            dictionary.Clear();
+            if (File.Exists(value))
```
AlanMorel, line 38: Maybe do an early return if the file doesn't exist by inverting this if statement

MapleServer2/Config/ConfigHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                file = value;
+                using (StreamReader sr = new StreamReader(file))
+                {
+                    string line, section = "";
```
AlanMorel, line 43: Can this be on 2 separate lines?

MapleServer2/Config/ConfigHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                if (instance == null)
+                {
+                    instance = new ConfigHandler("config.ini");
+                }
+                return instance;
+            }
+        }
+
+        public string commentDelimiter { get; set; }
```
AlanMorel, line 23: Does this need to be public?

AlanMorel (changes requested):

MapleServer2/Config/ConfigHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            if (value < minValue)
+            {
+                value = minValue;
+            }
+
+            if (value > maxValue)
+            {
+                value = maxValue;
+            }
```
AngeloTadeucci, line 137: Could just use Math.Min and Math.Min

MapleServer2/Config/ConfigHandler.cs
```diff
@@ -0,0 +1,199 @@
...
```
AngeloTadeucci, line 162: same thing here

AngeloTadeucci (changes requested):

PR #253, opened 2021-05-01
Fix: path for .env

AngeloTadeucci:


rhydev:
Going to PR a proper fix that uses Path variable.

PR #255, opened 2021-05-06
Database: Items, Visited Maps/Taxis, Trophies, Mails

AngeloTadeucci:
- Fixed `RequestTutorialItemHandler`;
- Moved char decorations (hair, eye, etc) to cosmetics;
- BCrypt hash passwords;
- Fixed VIP after relogging;
- Fixed taxi price;
- Fixed `ItemInventoryPacket`;
- Fixed `StorageInventory`, renamed to `StorageInventoryPacket`;
- Adjusted `ServerEnterPacket`;


MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
```diff
@@ -44,6 +44,7 @@ public override void Handle(GameSession session, PacketReader packet)
                     session.Player.Rotation = session.FieldPlayer.Rotation;
                     session.Player.Coord = session.Player.ReturnCoord;
                     session.Player.ReturnCoord.Z += Block.BLOCK_SIZE;
+                    DatabaseManager.UpdateCharacter(session.Player);
```
AlanMorel, line 47: just curious, any reason we need to save to database upon a map change?

MapleServer2/PacketHandlers/Login/LoginHandler.cs
```diff
@@ -38,17 +39,20 @@ public override void Handle(LoginSession session, PacketReader packet)
             string username = packet.ReadUnicodeString();
             string password = packet.ReadUnicodeString();
 
+            // Creates md5 hash of password
```
AlanMorel, line 42: No bcrypt in 2021?

MapleServer2/Types/Item.cs
```diff
@@ -186,5 +175,27 @@ public static bool IsPet(int itemId)
         {
             return itemId >= 60000001 && itemId < 61000000;
         }
+
+        public void GetMetadataValues(int id)
```
AlanMorel, line 179: this isn't really getting as much as it is setting, perhaps rename?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
AngeloTadeucci: We don't need to save on map change but I'm using as one way to manually test if changes are being saved. I could remove these, and add a command to save and implement a task to save every 10~20 minutes?

AlanMorel (changes requested):

Sparkymod (approved):

MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
AlanMorel: As long as this is the only place we are saving unnecessarily, then yeah it's fine

PR #256, opened 2021-05-07
Instrument Ensemble & Tick Changes

Zintixx:
## Instruments
- Created an `Instrument` class for handling instruments in a given field.
- Added `Instrument` to `FieldState` in order to have instruments be playing if a player enters a map mid-play.
- Ensembles now work because session server tick changes were made:

## Tick Changes
- a Player session is now properly having their ServerTick be set. Before, we were setting it with client tick.
- Normally, the client is supposed to request the ServerTick every 1 second from the server. It's not doing it (we're probably missing some packet or something). As a current workaround, I'm updating a session's ServerTick via `UserSyncHandler`

### Etc
- Added the default Emotes to the Player constructor

### DIFF
--- a/MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
@@ -105,6 +105,11 @@ namespace MapleServer2.PacketHandlers.Game
 
         private static void HandleStopImprovise(GameSession session)
         {
+            if (session.Player.Instrument == null)
+            {
+                return;
+            }
+
             session.FieldManager.BroadcastPacket(InstrumentPacket.StopImprovise(session.FieldPlayer));
             session.FieldManager.RemoveInstrument(session.Player.Instrument);
             session.Player.Instrument = null;
--- a/MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
@@ -200,10 +200,6 @@ namespace MapleServer2.PacketHandlers.Game
                 return;
             }
 
-            Item instrumentItem = session.Player.Inventory.Items[instrumentItemUid];
-
-            InsturmentInfoMetadata instrumentInfo = InstrumentInfoMetadataStorage.GetMetadata(instrumentItem.Function.Id);
-            InstrumentCategoryInfoMetadata instrumentCategory = InstrumentCategoryInfoMetadataStorage.GetMetadata(instrumentInfo.Category);
 
             Item score = session.Player.Inventory.Items[scoreItemUid];
 
@@ -212,6 +208,9 @@ namespace MapleServer2.PacketHandlers.Game
                 return;
             }
 
+            Item instrumentItem = session.Player.Inventory.Items[instrumentItemUid];
+            InsturmentInfoMetadata instrumentInfo = InstrumentInfoMetadataStorage.GetMetadata(instrumentItem.Function.Id);
+            InstrumentCategoryInfoMetadata instrumentCategory = InstrumentCategoryInfoMetadataStorage.GetMetadata(instrumentInfo.Category);
             Instrument instrument = new Instrument(instrumentCategory.GMId, instrumentCategory.PercussionId, score.IsCustomScore, session.FieldPlayer.ObjectId)
             {
                 Score = score,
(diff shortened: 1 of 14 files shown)


MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
```diff
@@ -150,12 +184,79 @@ private static void HandleCompose(GameSession session, PacketReader packet)
...
+            {
+                return;
+            }
+
+            if (!session.Player.Inventory.Items.ContainsKey(scoreItemUid) || !session.Player.Inventory.Items.ContainsKey(instrumentItemUid))
+            {
+                return;
+            }
+
+            Item instrumentItem = session.Player.Inventory.Items[instrumentItemUid];
```
AlanMorel, line 203: Lines 203-206 can be moved below 213 I think

AlanMorel (changes requested):

AlanMorel (approved):

MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
```diff
@@ -150,12 +184,84 @@ private static void HandleCompose(GameSession session, PacketReader packet)
...
+
+            session.Player.Instrument = session.FieldManager.RequestFieldObject(instrument);
+            session.Player.Instrument.Coord = session.FieldPlayer.Coord;
+
+            if (session.Player != party.Leader)
+            {
+                return;
+            }
+
+            int instrumentTick = session.ServerTick;
```
AngeloTadeucci, line 229: If the servertick changes when the loop is happening that wouldn't cause some desync?

MapleServer2/Types/FieldState.cs
```diff
@@ -16,6 +16,8 @@ public class FieldState
         public readonly ConcurrentDictionary<int, IFieldObject<GuideObject>> Guide;
         public readonly ConcurrentDictionary<int, IFieldObject<Cube>> Cubes;
         public readonly ConcurrentDictionary<int, IFieldObject<HealingSpot>> HealingSpots;
+        public readonly ConcurrentDictionary<int, IFieldObject<Instrument>> Instruments;
+
```
AngeloTadeucci, line 20: blank line

MapleServer2/Types/Player.cs
```diff
@@ -167,7 +168,7 @@ public Player(long accountId, long characterId, string name, byte gender, Job jo
             Titles = new List<int>();
             ChatSticker = new List<ChatSticker>();
             FavoriteStickers = new List<int>();
-            Emotes = new List<int>();
+            Emotes = new List<int>() { 90200011, 90200004, 90200024, 90200041, 90200042, 90200057, 90200043, 90200022, 90200031, 90200005, 90200006, 90200003, 90200092, 90200077, 90200073, 90200023, 90200001, 90200019, 90200020, 90200021 };
```
AngeloTadeucci, line 171: All new characters have these emotes?

MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
Zintixx: Yep! it would change as the loop is happening, which is why we're storing the current tick in `instrumentTick`

MapleServer2/Types/Player.cs
Zintixx: yeah, only about 4 of them are unlocked for new characters. The rest are locked behind trophies

AngeloTadeucci (approved):

MapleServer2/PacketHandlers/Game/InstrumentHandler.cs
AngeloTadeucci: im dumb

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -226,6 +226,18 @@ public void AddPlayer(GameSession sender, IFieldObject<Player> player)
                 sender.Send(RegionSkillPacket.Send(healingSpot.ObjectId, healingSpot.Value.Coord, new SkillCast(70000018, 1, 0, 1)));
             }
 
+            foreach (IFieldObject<Instrument> instrument in State.Instruments.Values)
+            {
+                if (instrument.Value.Improvise)
```
Sparkymod, line 231: You can use expression body on this too
```suggestion
                instrument.Value.Improvise == true ? sender.Send(InstrumentPacket.StartImprovise(instrument)) : sender.Send(InstrumentPacket.PlayScore(instrument));
```

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -309,6 +321,16 @@ public bool RemoveCube(IFieldObject<Cube> cube)
             return State.RemoveCube(cube.ObjectId);
         }
 
+        public void AddInstrument(IFieldObject<Instrument> instrument)
+        {
+            State.AddInstrument(instrument);
+        }
+
+        public bool RemoveInstrument(IFieldObject<Instrument> instrument)
```
Sparkymod, line 329: And return methods with a single line also can be express with lambda 
```suggestion
        public bool RemoveInstrument(IFieldObject<Instrument> instrument) => State.RemoveInstrument(instrument.ObjectId);
```

Sparkymod (approved):
The comments are tips, just to be more consistent in the code.

MapleServer2/Servers/Game/FieldManager.cs
darkvergus: i mean if you want to do return methods that only have a line to be expressed with lambda you will have alot to edit since there are quite a few.

PR #267, opened 2021-06-12
Feat: Gem Sockets and Coupon Use

Zintixx:
## Item Gem Sockets
- Pendants, Earrings, and Rings now generate gem sockets if the item level is 50 or above.
	- If the rarity is 3 (Blue), it will have 1 socket.
	- If the rarity is higher than 3, it will have 3 sockets
- 5% success rate for a gem socket to be unlocked upon item creation
- Crystal Fragment cost seems to be hardcoded into the client. I didn't find any indication that it was a packet/xml value

## Coupon Use
- Added a new opcode for `COUPON_USE`. This opcode takes care of unique functionality upon using coupons, such as inventory expansion, beauty vouchers, and character slots

## Character Slots
- Added CharacterSlots to `Account` to keep a track on how many character slots a player has

## Additional ItemUse entries
- Functional Character Slot coupons
- Functional Beauty Vouchers

## Beauty additions and Shop ID
- Added a variable for `ShopId` to the player. This will be stored in memory and a way to check and find what shop ID a player is in.
-  Beauty Handler no longer looks at the player's last Npctalk to get the shopId. This was needed due to beauty vouchers using similar handlers - some of which were finding the shopId via Npctalk. Because beauty vouchers involve no npc talk, this was revised.
- Added all entries for beauty voucher "shops"

## Misc addition
- Added a /commands command to display all current emulator commands

### Misc fixes
- Buying a different quantity on a meret market item is fixed
- Going to the event map via the UI button is fixed
	- Leaving a event map via Esc -> Leave actually brought up a different MoveField mode. Revised `MoveFieldHandler` to handle both.


MapleServer2/PacketHandlers/Game/ItemSocketSystemHandler.cs
```diff
@@ -30,32 +42,191 @@ public override void Handle(GameSession session, PacketReader packet)
...
+                if (item.Value.Amount >= crystalFragmentCost)
+                {
+                    InventoryController.Consume(session, item.Key, crystalFragmentCost);
+                    break;
+                }
+                else
+                {
+                    crystalFragmentCost -= item.Value.Amount;
+                    InventoryController.Consume(session, item.Key, item.Value.Amount);
+                }
```
AngeloTadeucci, line 143: Isn't line 141 useless? and can't you just use math.max?

MapleServer2/PacketHandlers/Game/ItemSocketSystemHandler.cs
```diff
@@ -67,6 +238,30 @@ private static void HandleUpgradeGem(GameSession session, PacketReader packet)
...
 
             ConsumeIngredients(session, metadata);
+
+            Item newGem = new Item(metadata.NextItemId)
+            {
+                IsLocked = gemstone.IsLocked,
+                UnlockTime = gemstone.UnlockTime,
+            };
+
+            Player oWner = GameServer.Storage.GetPlayerById(gemstone.OwnerId);
```
AngeloTadeucci, line 248: lowercase pls

MapleServer2/PacketHandlers/Game/ItemSocketSystemHandler.cs
```diff
@@ -111,16 +306,129 @@ private static void ConsumeIngredients(GameSession session, ItemGemstoneUpgradeM
...
+            Item gemstoneItem = new Item(gemstone.Id)
+            {
+                IsLocked = gemstone.IsLocked,
+                UnlockTime = gemstone.UnlockTime,
+                Rarity = 4
+            };
+
+            if (gemstone.OwnerId != 0)
+            {
+                Player oWner = GameServer.Storage.GetPlayerById(gemstone.OwnerId);
```
AngeloTadeucci, line 420: lowercase pls

AngeloTadeucci (changes requested):

MapleServer2/PacketHandlers/Game/ItemSocketSystemHandler.cs
Zintixx: I don't think line 141 is useless. It's needed to update the value of how many crystal fragments are left in the event that the item amount < cost.

GameDataParser/Parsers/BeautyParser.cs
```diff
@@ -5426,6 +5426,2590 @@ protected override List<BeautyMetadata> Parse()
...
+                ShopId = 601,
+                BeautyCategory = BeautyCategory.Standard,
+                BeautyType = BeautyShopType.Hair,
+                TokenType = ShopCurrencyType.Meso,
+                Items = new List<BeautyItem>
+                {
+                    new BeautyItem
+                    {
+                        ItemId = 10200103,
+                        Flag = ShopItemFlag.None,
```
AlanMorel, line 5440: Could make the None flag the default

MapleServer2/Packets/CharacterListPacket.cs
```diff
@@ -73,11 +73,11 @@ public static Packet DeleteCancel(long playerId)
             return pWriter;
         }
 
-        public static Packet SetMax(int unlocked, int total)
+        public static Packet SetMax(int unlocked)
```
AlanMorel, line 76: You could always leave the total parameter but make it default to 11

AlanMorel (changes requested):

AngeloTadeucci (approved):

MapleServer2/PacketHandlers/Game/ItemSocketSystemHandler.cs
AngeloTadeucci: `crystalFragmentCost` isn't used after the value is changed on line 141.  so it's useless 🤔

Zintixx: it's a loop tho so it is used. or am I misunderstanding?

AngeloTadeucci: welp im blind,

AlanMorel (approved):

PR #273, opened 2021-06-20
Add basic mob "AI"

quotopian:


MapleServer2/Tools/MobAIManager.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            MobAI ai = new MobAI();
+            foreach (XmlNode node in behaviorsNode)
+            {
+                MobAction newActionValue = GetMobAction(node.Attributes["action"]?.Value);
+                MobMovement movementValue = GetMobMovement(node.Attributes["movement"]?.Value);
+
+                if (node.Name == "spawn")
+                {
+                    ai.Rules.Add(MobState.Spawn, (newActionValue, movementValue, Array.Empty<MobAI.Condition>()));
```
AlanMorel, line 60: Can we make the empty array the default and initialize it in the constructor?

AlanMorel (changes requested):

MapleServer2/Tools/MobAIManager.cs
quotopian: I'm not sure what you mean?

The empty array here though is temporary. My idea in the AI files was that would be able to specify <condition>'s during a certain state to limit an action.

This part would parse a list of Conditions (lambdas) and the AI would run the conditions when there's a chance to take an action.

AngeloTadeucci (approved):
Think I reviewed most of the actual code, looking amazing!

AlanMorel (approved):

PR #274, opened 2021-06-20
Feat: FieldManager instances

AngeloTadeucci:
`ManagerFactory` is handling if _x_ `InstanceId` exists or not and then returning the correct `FieldManager` object.

Not sure if this is the best or correct way to implement instances, accepting any and all feedback.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
@@ -102,7 +102,7 @@ namespace MapleServer2.PacketHandlers.Game
                 }
             }
 
-            session.Player.Warp(coord, dstPortal.Rotation.ToFloat(), srcPortal.Target, 0);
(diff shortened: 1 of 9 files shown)


MapleServer2/PacketHandlers/Game/MoveFieldHandler.cs
```diff
@@ -102,7 +102,7 @@ public static void HandleMove(GameSession session, PacketReader packet)
                 }
             }
 
-            session.Player.Warp(coord, dstPortal.Rotation.ToFloat(), srcPortal.Target);
+            session.Player.Warp(coord, dstPortal.Rotation.ToFloat(), srcPortal.Target, 0);
```
AlanMorel, line 105: Can we make 0 the default?

AlanMorel (changes requested):

AlanMorel (approved):

PR #275, opened 2021-06-21
fix(vibrate): Coords and Tick Sync

integralcoder95:


MapleServer2/PacketHandlers/Game/VibrateHandler.cs
```diff
@@ -26,7 +26,7 @@ public override void Handle(GameSession session, PacketReader packet)
             int unk2 = packet.ReadInt();
             int unk3 = packet.ReadInt();
 
-            session.FieldManager.BroadcastPacket(VibratePacket.Vibrate(entityId, someId, objectId, flag, session.Player, session.ClientTick));
+            session.FieldManager.BroadcastPacket(VibratePacket.Vibrate(entityId, someId, objectId, flag, session.FieldPlayer, session.ServerTick));
```
AlanMorel, line 29: Why did the player need to become FieldPlayer?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/VibrateHandler.cs
integralcoder95: session.Player's coordinates only get updated on map enter/exit so using the field player coordinates is more accurate.

AngeloTadeucci (approved):

quotopian (approved):

PR #288, opened 2021-06-25
Fix: Moving between instances

AngeloTadeucci:
### Changed `ManagerFactory` to `FieldManagerFactory`

- Added **InstanceId** to `FieldManager`;
- Missing check if `FieldManager.InstanceId` is different from `Player.InstanceId`


AlanMorel (approved):

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -612,5 +613,15 @@ private Task StartSpawnTimer(IFieldObject<MobSpawn> mobSpawn)
...
+
+        public int Pin()
+        {
+            return ++Count;
+        }
+
+        public int Release()
+        {
+            return --Count;
+        }
```
AngeloTadeucci, line 625: Unsure about the naming of this attribute and functions, can't think of something better! Any suggestions?

MapleServer2/Servers/Game/FieldManager.cs
```diff
@@ -612,5 +613,15 @@ private Task StartSpawnTimer(IFieldObject<MobSpawn> mobSpawn)
                 }
             });
         }
+
+        public int Pin()
```
kOchirasu, line 617: You need to use `Interlocked.Increment` and `Interlocked.Decrement` for thread safety.

MapleServer2/Tools/FieldManagerFactory.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            lock (Managers)
+            {
+                if (!Managers.TryGetValue(key, out List<FieldManager> managerList))
+                {
+                    return false;
+                }
+
+                FieldManager fieldManager = managerList.FirstOrDefault(x => x.InstanceId == instanceId);
+                if (fieldManager == default || fieldManager.Release() > 0)
```
quotopian, line 48:
```suggestion
                if (fieldManager == default)
```

Why have the Factory decrement the player count? You could decrement the count inside `EnterField()` to separate concerns/duty. It would make it simpler to understand that Factory.Release() will always discard the field (and no side-effects like decrementing the player count).

Maybe the same idea with Pin()?

MapleServer2/Servers/Game/GameSession.cs
```diff
@@ -47,10 +44,10 @@ public void InitPlayer(Player player)
         public void EnterField(Player player)
         {
             // If moving maps, need to get the FieldManager for new map
-            if (player.MapId != FieldManager.MapId)
+            if (player.MapId != FieldManager.MapId || player.InstanceId != FieldManager.InstanceId)
             {
                 FieldManager.RemovePlayer(this, FieldPlayer); // Leave previous field
-                FieldManagerFactory.Release(FieldManager.MapId, player.InstanceId);
+                FieldManagerFactory.Release(FieldManager.MapId, FieldManager.InstanceId);
```
quotopian, line 50:
```suggestion
                if (FieldManager.Release() <= 0) {
                    FieldManagerFactory.Release(FieldManager.MapId, FieldManager.InstanceId);
                }
```

Goes with the other comment.

quotopian: Also, if you make this change, then you might need to lock over this entire block.

quotopian (changes requested):

AlanMorel (approved):

Sparkymod (approved):

PR #297, opened 2021-07-11
Feat: Additional Data to Skills. Parsing DurationTick ✔

Sparkymod:
## Skills ✨
- Parsing SubType and DurationTick on skills.
- Adding type `SkillAdditionalData` since this will have many more data to be add for status types, buff types, power ups on skills, splash effects, etc..

Open to suggestions 😊


AngeloTadeucci (approved):

darkvergus (approved):

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -115,6 +116,29 @@ protected override List<SkillMetadata> Parse()
                     }
                 }
             }
+            // Parsing Additional Data
+            foreach (PackFileEntry entry in Resources.XmlFiles)
+            {
+                if (entry.Name.StartsWith("additionaleffect"))
```
AlanMorel, line 122: You can invert and continue

AlanMorel (changes requested):

PR #292, opened 2021-07-02
Dungeon lobby system

Clockworkx:
Should be ready to be merged now. 

### Dungeon Creation and Entrance

- Handled for Solo and Groups.

### Misc

- Added First triggers, see SetPortal, ShowGuideSummary and CheckUserCount for examples.
- Changed Player.Warp to calculate a random spawn by itself, but can overwrite the random spawn/rotation by passing those to the function. (see function declaration)
- Oneshot command for testing mob kills.
- Parsing all portal types
- Dungeon Parser

PR #298, opened 2021-07-16
Fix: tutorial barrier portal

AngeloTadeucci:
With the recent PR #292 it was implemented an trigger function **SetPortal** which is called when entering the tutorial map that sets all portals to invisible and disabled. 

This is an temporary fix so the portal appears after breaking the barrier.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/SkillHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/SkillHandler.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using Maple2Storage.Types;
 using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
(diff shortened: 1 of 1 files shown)


MapleServer2/PacketHandlers/Game/SkillHandler.cs
```diff
@@ -163,6 +164,10 @@ private static void HandleAoeDamage(GameSession session, PacketReader packet)
                         session.Send("4F 00 03 E8 03 00 00 00 00 00 00 00 00 00 00 00 00 80 3F".ToByteArray());
                         session.Send("4F 00 03 D0 07 00 00 00 00 00 00 00 00 00 00 00 00 80 3F".ToByteArray());
                         session.Send("4F 00 08 01 04 01 00 00".ToByteArray());
+
```
Clockworkx, line 167: Have a comment that explains the purpose of this "temp fix barrier portal" or something

AngeloTadeucci: its on line 162 actually :P

AlanMorel (approved):

MapleServer2/PacketHandlers/Game/SkillHandler.cs
Clockworkx: my bad :D

Clockworkx (approved):

Clockworkx (approved):

Sparkymod (approved):

PR #306, opened 2021-07-22
Fixing MapMetadata, GameCommandActions & Others

AngeloTadeucci:
### MapMetadata, MapParser and MapMetadataStorage

- Changed **Blocks** from `List<MapBlock>` to `Dictionary<CoordS, MapBlock>`.

### GameCommandActions

- New command `/goto playerName`;
- Fixed `/coord` to use **UserMoveByPortalPacket**;
- Added bool to **ParseShort**, **ParseInt** and **ParseLong** to return if it succeed or not.

### Others

- Adding null checks for **MapMetadata**, **ItemEquipHandler** and **FieldManager**;
- Removed logger prints from character creation;
- Ignoring Player.PartyId from database.


MapleServer2/Data/Static/MapMetadataStorage.cs
```diff
@@ -30,26 +29,35 @@ public static MapMetadata GetMetadata(int mapId)
         public static bool BlockExists(int mapId, CoordS coord)
         {
             MapMetadata mapD = GetMetadata(mapId);
```
AlanMorel, line 31: I know this was here before this PR but can this be renamed to `mapMetadata`, `mapD` is a bad name for this

MapleServer2/Tools/GameCommandActions.cs
```diff
@@ -109,9 +168,44 @@ public static void Process(GameSession session, string command)
...
+            {
+                session.Send(UserMoveByPortalPacket.Move(session.FieldPlayer.ObjectId, fieldPlayer.Coord, fieldPlayer.Rotation));
+                return;
+            }
+            session.Player.Warp(mapId: target.MapId, coord: fieldPlayer.Coord, instanceId: target.InstanceId);
+        }
+
+        private static void ProcessOneshotCommand(GameSession session)
+        {
+            Player player = session.Player;
```
AlanMorel, line 191: Could get rid of this variable since it's not being used on line 199 when it could be

AlanMorel (changes requested):

AlanMorel (approved):

Sparkymod (approved):

PR #334, opened 2021-08-13
Fix: Beauty crash and returning to correct map

AngeloTadeucci:
Fixes crash when changing eyes/hair for the second time;
Only setting return map if character isn't inside an beauty map.

### DIFF
--- a/MapleServer2/PacketHandlers/Game/BeautyHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/BeautyHandler.cs
@@ -356,7 +356,7 @@ namespace MapleServer2.PacketHandlers.Game
                     return;
             }
 
-            if (session.Player.MapId is not (int) Map.RosettaBeautySalon or (int) Map.TriaPlasticSurgery or (int) Map.DouglasDyeWorkshop)
+            if (MapEntityStorage.HasSafePortal(session.Player.MapId))
             {
                 session.Player.ReturnCoord = session.FieldPlayer.Coord;
                 session.Player.ReturnMapId = session.Player.MapId;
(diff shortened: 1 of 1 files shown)


MapleServer2/PacketHandlers/Game/BeautyHandler.cs
```diff
@@ -360,8 +356,11 @@ private static void HandleTeleport(GameSession session, PacketReader packet)
                     return;
             }
 
-            session.Player.ReturnCoord = session.FieldPlayer.Coord;
-            session.Player.ReturnMapId = session.Player.MapId;
+            if (session.Player.MapId is not (int) Map.RosettaBeautySalon or (int) Map.TriaPlasticSurgery or (int) Map.DouglasDyeWorkshop)
+            {
+                session.Player.ReturnCoord = session.FieldPlayer.Coord;
+                session.Player.ReturnMapId = session.Player.MapId;
+            }
```
Zintixx, line 363: Hmm this might be a bigger problem if for some reason the return map and coord were being changed for an instance map...Normally all instanced maps should not save return data.

AngeloTadeucci: I changed this because if you were inside Rosetta and then went to Plastic Surgery you went back to rosetta and kept going back to rosetta. You think we should save the instance where the player left? Any examples you have in mind?

AlanMorel (approved):

Sparkymod (approved):

MapleServer2/PacketHandlers/Game/BeautyHandler.cs
Zintixx: Sorry mixing up instance with maps with no safe portals. Basically if the map the player is entering has no safe portals, their return coord/maps shouldn't be saved. Which is what you did here for the beauty maps...I'm thinking there will be several other examples where would happen. Dungeons, some tutorial/quest maps, etc

Zintixx (approved):

PR #340, opened 2021-08-22
Feat: database v2

AngeloTadeucci:
Removed **EntityFramework**, now using SQL with [SQLKata ](https://sqlkata.com/) package as an query builder.


MapleServer2/Database/Classes/DatabaseAccount.cs
```diff
@@ -0,0 +1,125 @@
+﻿using Maple2Storage.Types;
+using MapleServer2.Enums;
+using MapleServer2.Types;
+using Newtonsoft.Json;
+using SqlKata.Execution;
+
+namespace MapleServer2.Database.Classes
+{
+    public static class DatabaseAccount
```
AlanMorel, line 9: Rather than calling this class `DatabaseAccount`, can it instead be called `Accounts` and not static, then in a file called `Database` create an instance of `Accounts` then expose it using a getter? So the ultimate caller uses something like `Database.getAccounts().Create()` or `Database.getAccounts().FindById()`

MapleServer2/Database/Classes/DatabaseAccount.cs
```diff
@@ -0,0 +1,125 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        internal static List<Player> FindAllByAccountId(long accountId)
+        {
+            List<Player> characters = new List<Player>();
+
+            IEnumerable<dynamic> result = DatabaseManager.QueryFactory.Query("Characters").Where("AccountId", accountId)
```
AlanMorel, line 53: Maybe querying the characters table should not happen in the `Account` file?

AlanMorel (changes requested):

MapleServer2/Database/Classes/DatabaseAccount.cs
AngeloTadeucci: oh 100% my bad, it was one of the first stuff I worked on

AlanMorel (approved):

darkvergus (approved):

Sparkymod (approved):

PR #352, opened 2021-08-31
Tutorial Item and Default Items check

Zintixx:
Parsing `defaultitems.xml` to check for default items. No more XML editing your starter gear.
Added a (proper) check to giving the player a tutorial Item only if they're missing it from their inventory.
Fixed `SetRandomMesh` to now work properly. Sole Survivor now working as intended.


MapleServer2/Data/Static/DefaultItemsMetadataStorage.cs
```diff
@@ -0,0 +1,31 @@
...
+                defaultItems[item.JobCode] = item;
+            }
+        }
+
+        public static bool IsValid(int job, int itemId)
+        {
+            DefaultItemsMetadata metadata = defaultItems.GetValueOrDefault(job);
+            if (!metadata.DefaultItems.Any(x => x.ItemId == itemId))
+            {
+                return defaultItems.GetValueOrDefault(0).DefaultItems.Any(x => x.ItemId == itemId);
```
AngeloTadeucci, line 26: This looks a bit confusing because of the naming, maybe rename the dictionary?

AngeloTadeucci (approved):
LGTM!

Sparkymod (commented):
Looks good! Just a question, instead of making a new `DefaultItemsMetadataStorage` you can use what we currently have of retrieving items of a storage `ItemsMetadataStorage` you just need to set the initializer in the same constructor, no?

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
```diff
@@ -90,29 +99,34 @@ public static void HandleCreate(LoginSession session, PacketReader packet)
...
                         // Hair Length/Position
-                        float backLength = BitConverter.ToSingle(packet.Read(4), 0);
+                        float backLength = packet.ReadFloat();
                         CoordF backPositionCoord = packet.Read<CoordF>();
                         CoordF backPositionRotation = packet.Read<CoordF>();
-                        float frontLength = BitConverter.ToSingle(packet.Read(4), 0);
+                        float frontLength = packet.ReadFloat();
                         CoordF frontPositionCoord = packet.Read<CoordF>();
                         CoordF frontPositionRotation = packet.Read<CoordF>();
+                        if (!DefaultItemsMetadataStorage.IsValid((int) job, id))
```
AlanMorel, line 124: Can't this check just be done after line 105?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
Zintixx: No because depending on the item slot type, there's more packets to read. If this was done after line 105, it would jumble everything up.

Zintixx:
> Looks good! Just a question, instead of making a new `DefaultItemsMetadataStorage` you can use what we currently have of retrieving items of a storage `ItemsMetadataStorage` you just need to set the initializer in the same constructor, no?

I'm not sure what you mean? ItemsMetadataStorage doesn't have a list/dictionary of default items. I can't check with ItemsMetadataStorage if they're items from character creation.

MapleServer2/PacketHandlers/Login/CharacterManagementHandler.cs
AlanMorel: Yeah that makes sense, good point

AlanMorel (approved):

PR #355, opened 2021-08-31
Fix: Database Shop

AngeloTadeucci:
Return null if shop doesn't exists.

### DIFF
--- a/MapleServer2/Database/Classes/DatabaseShop.cs
+++ b/MapleServer2/Database/Classes/DatabaseShop.cs
@@ -10,6 +10,10 @@ namespace MapleServer2.Database.Classes
         public Shop FindById(long id)
         {
             Shop shop = QueryFactory.Query(TableName).Where("Id", id).Get<Shop>().FirstOrDefault();
+            if (shop == default)
+            {
+                return null;
+            }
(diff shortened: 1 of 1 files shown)


MapleServer2/Database/Classes/DatabaseShop.cs
```diff
@@ -10,6 +10,10 @@ public class DatabaseShop : DatabaseTable
         public Shop FindById(long id)
         {
             Shop shop = QueryFactory.Query(TableName).Where("Id", id).Get<Shop>().FirstOrDefault();
```
Sparkymod, line 12: you can use `.First()` to return a null if not found. i think you can.
```suggestion
            Shop shop = QueryFactory.Query(TableName).Where("Id", id).Get<Shop>().First();
```

Sparkymod (changes requested):

MapleServer2/Database/Classes/DatabaseShop.cs
AngeloTadeucci: With first it will return ArgumentNullException or InvalidOperationException

Sparkymod (approved):

PR #358, opened 2021-09-02
Fix: Improve the load of all Metadata.

Sparkymod:
- Pre-Load all metadata when the server starts.

### DIFF
--- a/MapleServer2/MapleServer.cs
+++ b/MapleServer2/MapleServer.cs
@@ -3,13 +3,13 @@ using Autofac;
 using MaplePacketLib2.Tools;
 using MapleServer2.Constants;
 using MapleServer2.Database;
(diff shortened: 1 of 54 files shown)


MapleServer2/MapleServer.cs
```diff
@@ -53,6 +54,8 @@ public static void Main(string[] args)
             GameServer = gameScope.Resolve<GameServer>();
             GameServer.Start();
 
+            Logger.Info("Server Started.".Pastel("#aced66"));
```
AlanMorel, line 57: Can we have a single location for hex colors rather than hardcoding it here?

AlanMorel (changes requested):

AngeloTadeucci (approved):

PR #374, opened 2021-09-07
Fix: FallDamageHandler missing packet send

Sparkymod:
- Fix an issue where it wasn't showing the UI of beauty with mounts


AngeloTadeucci (approved):

MapleServer2/PacketHandlers/Game/NpcTalkHandler.cs
```diff
@@ -222,6 +222,7 @@ private static void HandleBeauty(GameSession session)
                     break;
             }
             session.Send(UserMoveByPortalPacket.Move(session.FieldPlayer, portal.Coord.ToFloat(), portal.Rotation.ToFloat()));
+            session.FieldManager.BroadcastPacket(MountPacket.StopRide(session.FieldPlayer, false));
```
Zintixx, line 225: While this would fix it, I feel like there may be more testing involved. We need to see if every time the packet `UserMoveByPortalPacket.Move` is sent, dismounting should always follow. If so, it might be better to just make this as a method to use to bundle it together everytime.

PR #377, opened 2021-09-09
Minor Fixes

darkvergus:
- Fix changing the name of database causing it to crash, Static until we find a better solution.
- Save on crash.
- Minor change to logger in Metadata


.env.example
```diff
@@ -1,7 +1,6 @@
 # Database Info
 DB_IP=localhost
 DB_PORT=3306
-DB_NAME=Maple2DB
```
AngeloTadeucci, line 4: Fix `.env.example` too

AngeloTadeucci (changes requested):

AngeloTadeucci (approved):

.env.example
AngeloTadeucci: im dumb :D

Issue #322, opened 2021-07-29
SoulBinder can't basic attack with their orb equipped


Snowawa:
When you try to basic attack with an orb equip it'll say "You have not learned this skill yet."

PR #378, opened 2021-09-10
Fix: SkillParser & RegionSkills

Sparkymod:
Fix: #322 
- Now all skills are being parse correctly.
- Adding more skill type checks.
- Adding GM skills type.
- Now Skill Region are being cast correctly. (WIP)
- Correct implementation of skill Levels.
- Adding `Buff` and `Skill` commands
- Minor other fixes.


MapleServer2/Commands/Game/PlayerCommands.cs
```diff
@@ -127,4 +132,75 @@ public override void Execute(GameCommandTrigger trigger)
...
+    public class BuffCommand : InGameCommand
+    {
+        public BuffCommand()
+        {
+            Aliases = new[]
+{
+                "buff"
+            };
+            Description = "Level up all the skills available.";
+            AddParameter<int>("id", "ID of the status.");
```
AlanMorel, line 175: Unrelated to this PR but I think parameters should be an array on the Command class, and then you initialize it like this: 
```
Parameters = new [] {
    Parameter<int>("id", "ID"),
    Parameter<int>("id", "ID")
}
```

MapleServer2/Servers/Game/GameSession.cs
```diff
@@ -26,9 +26,9 @@ public GameSession(FieldManagerFactory fieldManagerFactory) : base()
             FieldManagerFactory = fieldManagerFactory;
         }
 
-        public void SendNotice(string message)
+        public void SendNotice(object message)
```
AlanMorel, line 29: Why?

AlanMorel (changes requested):

MapleServer2/Servers/Game/GameSession.cs
Sparkymod: Is better.

1. If you just one to send an object dont need to add `ToString()` or do interpolation string just for an object is easier to just pass the string or the object itself.

2.Easy to understand since you are `Sending a Notice about something`.

MapleServer2/Commands/Game/PlayerCommands.cs
Sparkymod: Thats a good idea!

AngeloTadeucci (approved):

GameDataParser/Parsers/SkillParser.cs
```diff
@@ -88,55 +102,38 @@ protected override List<SkillMetadata> Parse()
...
                     foreach (XmlNode job in jobs)
                     {
-                        if (job.Attributes["feature"] != null) // Getting attribute that just have "feature"
+                        // Grabs all the skills and them the jobCode.
+                        XmlNodeList skills = job.SelectNodes("skills/skill");
+                        int jobCode = int.Parse(job.Attributes["code"].Value);
+                        foreach (XmlNode skill in skills)
                         {
-                            string feature = job.Attributes["feature"].Value;
-                            if (feature == "JobChange_02") // Getting JobChange_02 skillList for now until better handle Awakening system.
```
Zintixx, line 94: Why was this removed?

Sparkymod: What this was use for is to set just the `feature =JobChange_02` skills in the job.xml, and with the new changes we are setting all skills and subskills to their job code, including beginner.

darkvergus (approved):
POGGERS

Zintixx (approved):

PR #409, opened 2021-09-18
Feat: Skills path

Sparkymod:
- Parse magicPath for skills.
- Fix skill region cast to field(now most of them show their tiles affected)
- Fix region skills not being remove after first cast.


AngeloTadeucci (approved):
LGTM

darkvergus (approved):

PR #412, opened 2021-09-19
Fix creating home

darkvergus:
- Fix creating home with an empty template


AngeloTadeucci (approved):

Zintixx (commented):
Don't users who pick no home template still get some blocks to start off with?

rhydev:
Empty template just has the base ground blocks, the blocks that you can't place yourself idk what to call them.

PR #428, opened 2021-09-29
feat: Item Linking

Clockworkx:
### Item Linking
implemented

### Misc
change of shop sell packet as it was creating an unnecessary item

### DIFF
--- a/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
@@ -256,7 +256,6 @@ namespace MapleServer2.PacketHandlers.Game
                     {
                         item = DatabaseManager.Items.FindByUid(itemUid);
                     }
-                    break;
                 }
                 if (item != null)
                 {
--- a/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/UserChatHandler.cs
@@ -1,6 +1,8 @@
-﻿using MaplePacketLib2.Tools;
+﻿using System.Text.RegularExpressions;
+using MaplePacketLib2.Tools;
 using MapleServer2.Commands.Core;
 using MapleServer2.Constants;
+using MapleServer2.Database;
 using MapleServer2.Enums;
 using MapleServer2.Managers;
 using MapleServer2.Packets;
@@ -33,34 +35,58 @@ namespace MapleServer2.PacketHandlers.Game
                 return;
             }
 
+            Packet itemLinkPacket = null;
+
+            // '<' signals a message containing an item link
+            if (message.Contains("<A"))
+            {
+                string[] itemLinkMessages = Regex.Matches(message, @"<A (.*?)>")
+                    .Cast<Match>()
+                    .Select(m => m.Value)
+                    .ToArray();
+                List<Item> item = new();
+
+                foreach (string itemLinkMessage in itemLinkMessages)
+                {
+                    string[] itemLinkMessageSplit = itemLinkMessage.Split(',');
+                    long itemUid = long.Parse(itemLinkMessageSplit[1]);
+
+                    item.Add(DatabaseManager.Items.FindByUid(itemUid));
+                    if (item != null)
+                    {
+                        itemLinkPacket = ItemLinkPacket.SendLinkItem(item);
+                    }
+                }
+            }
+
             switch (type)
             {
                 case ChatType.Channel:
                     HandleChannelChat(/*session, message, type*/);
                     break;
                 case ChatType.Super:
-                    HandleSuperChat(session, message, type);
+                    HandleSuperChat(session, message, type, itemLinkPacket);
                     break;
                 case ChatType.World:
-                    HandleWorldChat(session, message, type);
+                    HandleWorldChat(session, message, type, itemLinkPacket);
                     break;
                 case ChatType.GuildAlert:
-                    HandleGuildAlert(session, message, type);
+                    HandleGuildAlert(session, message, type, itemLinkPacket);
                     break;
                 case ChatType.Guild:
-                    HandleGuildChat(session, message, type);
+                    HandleGuildChat(session, message, type, itemLinkPacket);
                     break;
                 case ChatType.Party:
-                    HandlePartyChat(session, message, type);
+                    HandlePartyChat(session, message, type, itemLinkPacket);
                     break;
                 case ChatType.WhisperTo:
-                    HandleWhisperChat(session, recipient, message);
+                    HandleWhisperChat(session, recipient, message, itemLinkPacket);
                     break;
                 case ChatType.Club:
-                    HandleClubChat(/*session, message, type, clubId*/);
+                    HandleClubChat(/*session, message, type, clubId, itemLinkPacket*/);
                     break;
                 default:
-                    HandleChat(session, message, type);
+                    HandleChat(session, message, type, itemLinkPacket);
                     break;
             }
         }
@@ -71,7 +97,7 @@ namespace MapleServer2.PacketHandlers.Game
             // session.Send(NoticePacket.Notice(SystemNotice.UsedChannelChatVoucher, NoticeType.ChatAndFastText));
         }
 
-        private static void HandleSuperChat(GameSession session, string message, ChatType type)
+        private static void HandleSuperChat(GameSession session, string message, ChatType type, Packet itemLinkPacket)
         {
             if (session.Player.SuperChat == 0)
             {
@@ -87,13 +113,17 @@ namespace MapleServer2.PacketHandlers.Game
                 return;
             }
 
+            if (itemLinkPacket != null)
+            {
+                MapleServer.BroadcastPacketAll(itemLinkPacket);
+            }
             MapleServer.BroadcastPacketAll(ChatPacket.Send(session.Player, message, type));
             InventoryController.Consume(session, superChatItem.Uid, 1);
             session.Send(SuperChatPacket.Deselect(session.FieldPlayer));
             session.Player.SuperChat = 0;
         }
 
-        private static void HandleWorldChat(GameSession session, string message, ChatType type)
+        private static void HandleWorldChat(GameSession session, string message, ChatType type, Packet itemLinkPacket)
         {
             Item voucher = session.Player.Inventory.Items.Values.FirstOrDefault(x => x.Tag == "FreeWorldChatCoupon");
             if (voucher != null)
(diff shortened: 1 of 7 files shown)


MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -33,34 +35,58 @@ public override void Handle(GameSession session, PacketReader packet)
...
+                    .ToArray();
+                List<Item> item = new();
+
+                foreach (string itemLinkMessage in itemLinkMessages)
+                {
+                    string[] itemLinkMessageSplit = itemLinkMessage.Split(',');
+                    long itemUid = long.Parse(itemLinkMessageSplit[1]);
+
+                    item.Add(DatabaseManager.Items.FindByUid(itemUid));
+                    if (item != null)
```
AngeloTadeucci, line 55: Item will always be not  null also change it to check for the count and you need to move it outside the for each.

AngeloTadeucci (changes requested):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
Clockworkx: should be good now, thanks

AngeloTadeucci (approved):

darkvergus (approved):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -33,34 +35,62 @@ public override void Handle(GameSession session, PacketReader packet)
                 return;
             }
 
+            Packet itemLinkPacket = null;
```
AlanMorel, line 38: Can all of this be put in a separate method that takes in a message and returns `itemLinkPacket` or null?

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
Clockworkx: if you like that better? sure

Clockworkx: oh wait you mean the whole item linking part in a function?
I don't see the benefit of it, since it is only used once

AlanMorel: Yeah it's only being used once but it's an isolated functionality, getting the item link packet, it would read better in a separate function even if that function will only be called once

Clockworkx: yea reading is for sure better, i'm always just concerned about the overhead of the functioon call.
But if you think it's worth it, i'll certainly do it

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -169,18 +194,82 @@ private static void HandleWhisperChat(GameSession session, string recipient, str
...
+        private static Packet GetItemLink(string message)
+        {
+            // '<' signals a message containing an item link
+            if (!message.Contains('<'))
+            {
+                return null;
+            }
+            Packet itemLinkPacket = null;
+
+            Match a = Regex.Match(message, @"<A (.*?)>");
```
AngeloTadeucci, line 229: Not being used?

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -169,18 +194,82 @@ private static void HandleWhisperChat(GameSession session, string recipient, str
...
+            foreach (string itemLinkMessage in itemLinkMessages)
+            {
+                string[] itemLinkMessageSplit = itemLinkMessage.Split(',');
+                string itemLinkType = Regex.Match(itemLinkMessageSplit[0], @"(?<=:)(.*)").Value;
+
+                long itemUid = long.Parse(itemLinkMessageSplit[1]);
+                Item item = null;
+                switch (itemLinkType)
+                {
+                    case "itemTooltip":
```
AngeloTadeucci, line 245: switch with only one case? It's supposed to be more cases? If so, add an comment

MapleServer2/Packets/MeretMarketPacket.cs
```diff
@@ -118,17 +118,17 @@ public static Packet Promos(List<MeretMarketItem> items)
...
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
+            pWriter.WriteByte();
```
AngeloTadeucci, line 131: Maybe use `pWriter.WriteZero(10)`?

AngeloTadeucci (changes requested):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -169,18 +194,82 @@ private static void HandleWhisperChat(GameSession session, string recipient, str
...
+        {
+            // '<' signals a message containing an item link
+            if (!message.Contains('<'))
+            {
+                return null;
+            }
+            Packet itemLinkPacket = null;
+
+            Match a = Regex.Match(message, @"<A (.*?)>");
+            string[] itemLinkMessages = Regex.Matches(message, @"<A (.*?)>")
```
AlanMorel, line 230: Since it appears that HTML is being parsed here, you should instead use an HTML or XML parser. The emulator already makes use of the latter

AlanMorel (changes requested):

MapleServer2/Packets/MeretMarketPacket.cs
Zintixx: I'm opposed to this because it doesn't indicate the packet structure. These single bytes might mean something we don't know yet.

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
Clockworkx: yea I forgot about it

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
Clockworkx: the only part being html here is that the link is enclosed in a html <a> tag. 
the <a is not relevant by itself but it ensures people can't fake item links via chat messages, as the < is escaped by the client if typed in chat.
I don't want to spend more time on item linking, it works and I'm not sure what the benefit would be using html parser here

Clockworkx: I guess I could use the htm parser for the one line in question here, but I don't know how to use it and I don't have energy left for item linking.

AngeloTadeucci (approved):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
AlanMorel: Not sure why you insist on being hesitant to improve your code, but the benefits are that you can dramatically improve readability by using a parser. Sure, your code works, but code simply working shouldn't be the minimum acceptance criteria. I legitimately could not tell what your code was trying to parse, especially on line 238, and I had to ask in Discord for an example. 

Here's some sample code that is shorter, much more readable (even if you had no idea what the input looked like, this code basically documents what it is trying to parse for you), and more robust since we are offloading the parsing to a battle-tested native library rather than writing custom regex:

```
        private static void GetItemLink(string message)
        {
            XmlDocument doc = new XmlDocument();
            doc.LoadXml("<xml>" + message + "</xml>");
            foreach (XmlNode link in doc.SelectNodes("//A"))
            {
                string[] hrefSplit = link.Attributes["HREF"].Value.Split(",");
                string itemLinkType = hrefSplit[0].Split(":")[1];
                long itemUid = long.Parse(hrefSplit[1]);
            }
        }
```
Parsing what is clearly HTML/XML using regex is universally considered a bad idea for many reasons.

AlanMorel (changes requested):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
darkvergus: gotta agree with alan using the html parser or even the xml parser which already exists is a better option. Also we might only have this example that users html tags for now but in the future we might find more.

Clockworkx: The code didn't work as is and I am not sure of what is going wrong. 
My main point is I want to move on to another feature. 
I think my code is fine and readable. 
If someone looks at the variable names it's very understandable what's going on.
Since the system here has multiple layers to get the information of the item link, of course one needs to take the time to understand each layer (in TERMS of the variables  that are being created).
Trying to understand exactly what each string manipulation is doing is completely unnecessary.
It's like someone would look at every functioon used in another function in order to understand what is going on in the function those functions are called in.
The string manipulations do not need to be understood in order to understand the code, and there are as many instructions with the parser, actually even more when creating the parser instance. Therefore I think they are equally understandable.
All someone needs to understand is, "a regex is being parsed here and assigned to this variable"
no need to understand the content of the regex.
Don't get me wrong I really appreciate the input from you, I learned a lot by that.

While this code is html syntax we don't care that is html, we don't need any of the html information.
We could use another part of the string to match with, it would also work.
I only do it to avoid people creating a fake item message by typing the syntax of the item link in chat, as the server gets the item link info from the message and there is no mode or other information which makes the item link message unique. The only thing that makes it unique is the "<" character, which is escaped by the client to "&lt" so that is the only way to make sure it was send by the client.

A regex should be more performant here as well, no need of all the class overhead of the parser.
It is in my opinion as readable as the other.

I am not resistant to improving my code, but I don't want to make that many revisions, when the change is so simple and as I tried to explain, in my perspective, has little benefit if at all. 

and while reading html by regexs would probably be bad in html contexts, I think "reading html" is the problem here, we could as well read from : to , by a regex or use other string manipulation. 

Lastly, I want to move on to other features and not spend 2 days of my free time on 25 lines of code (the research took the most time of course - and testing) The code works, and I think the syntax is understandable. 
I tried changing it to the xml parser EVEN though I dislike it compared to my code, but it gave me errors.
So, if someone wants to change it to using an xml/html parser/, they can do it, although I think it's unneccessary. I will continue with another feature or so from now on.

Sparkymod: While arbitrary HTML with just a regex is a bad idea and almost impossible, it's sometimes appropriate to use them for parsing a limited, known set of HTML. In this case, you have a small set of HTML tags that you want to extract data from the package, regular expressions can work fine. Sometimes we have to know the scope to be used, and this is just a particular case.

Sparkymod (approved):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -169,18 +195,79 @@ private static void HandleWhisperChat(GameSession session, string recipient, str
...
+                            {
+                                Uid = itemUid
+                            };
+                        }
+                    }
+                    else if (itemToolTipType == 3) // normal item
+                    {
+                        item = DatabaseManager.Items.FindByUid(itemUid);
+                    }
+                    break;
```
AngeloTadeucci, line 259:
```suggestion
```

AngeloTadeucci (approved):

MapleServer2/PacketHandlers/Game/UserChatHandler.cs
```diff
@@ -1,6 +1,10 @@
-﻿using MaplePacketLib2.Tools;
+﻿using System.Text.RegularExpressions;
```
AngeloTadeucci, line 1:
```suggestion
```

PR #435, opened 2021-10-20
GameDataParser cleanup

AngeloTadeucci:
Clean up of GameDataParser:

- Simpler null checks;
- Simpler parsing all values of arrays using LINQ;
- Removed some cases of empty objects being parsed;
- Added `throw new Exception` if task fails, before it wouldn't alert that one task failed because of the progress bar.

I'm not afraid of reverting some changes if necessary. So give all feedback please :)


AlanMorel (approved):

GameDataParser/Parsers/ItemOptionStaticParser.cs
```diff
@@ -26,95 +26,48 @@ protected override List<ItemOptionStaticMetadata> Parse()
...
+                                optionsStatic.Stats.Add(new ParserStat(ItemAttribute.CriticalDamage, int.Parse(node.Attributes[item.Name].Value.Split(",").First())));
                                 break;
                             case "cap_value_base":
-                                if (node.Attributes[item.Name].Value.Contains(','))
-                                {
-                                    optionsStatic.Stats.Add(new ParserStat(ItemAttribute.CriticalRate, int.Parse(node.Attributes[item.Name].Value.Split(",").First())));
-                                    break;
-                                }
-                                optionsStatic.Stats.Add(new ParserStat(ItemAttribute.CriticalRate, int.Parse(node.Attributes[item.Name].Value)));
+                                optionsStatic.Stats.Add(new ParserStat(ItemAttribute.CriticalRate, int.Parse(node.Attributes[item.Name].Value.Split(",").First())));
```
Sparkymod, line 58: What happen then if is not contains the ',' ?

AlanMorel: The entire thing becomes a 1-element array

Sparkymod (approved):

PR #448, opened 2021-10-31
Fix: Fast Travel

### LINKED ISSUE #445: Using fast travel with helicopter with VIP consumes mesos
When you have VIP the client uses the meso function but doesn't check if the character has VIP. 
With VIP using fast travel with heli is free.

Snowawa:
Fixes #445


Clockworkx:
I would change the comment to something like
"VIPs can travel without cost"
"VIP Travel"

The scriptloader part is not needed

AlanMorel (approved):

AngeloTadeucci (approved):
Change the comment like Saint said and it's perfect

Zintixx (approved):

Sparkymod (approved):

PR #455, opened 2021-11-02
fix: add check for mysql version

### LINKED ISSUE #452: Check for MySQL version on server start
Some code to help you get working:

Inside `InitDatabase()` in `MapleServer.cs` before 
```
if (DatabaseManager.DatabaseExists())
{
    Logger.Info("Database already exists.");
    return;
}
```
 just do something like 
```
if (DatabaseManager.getVersion() < MINIMUM_VERSION_HERE) {
    Logger.Error("MySQL version out-of-date, please upgrade to version " + MINIMUM_VERSION_HERE + ".");
    return;
}
```

You can check the MySQL version with something like this:
```
string connString = "server=localhost;userid=root;password=root;database=mydb";
MySqlConnection conn = new MySqlConnection(connString);

conn.Open();
Console.WriteLine("MySQL version : " + conn.ServerVersion);
```

Result:
```
MySQL version : 8.0.22
Press any key to continue . . .
```
Split the version string by a `.`, take the first number, cast it to an int, compare it with a predefined default minimum version and throw an exception if it's lower.

michaelowens:
Fixes #452 

Remaks:
- Is there a central place to put the required version number? Currently I put it in `InitDatabase`
- The original issue suggested returning from `InitDatabase` when the check fails. This does not stop the emulator from running. Should a shutdown be enforced or just let it crash and possibly burn?

### DIFF
--- a/MapleServer2/Database/DatabaseManager.cs
+++ b/MapleServer2/Database/DatabaseManager.cs
@@ -56,6 +56,14 @@ namespace MapleServer2.Database
 
         public static void RunQuery(string query) => new QueryFactory(new MySqlConnection(ConnectionString), new MySqlCompiler()).Statement(query);
 
+        public static int getVersion()
+        {
+            MySqlConnection conn = new MySqlConnection(ConnectionString);
+            conn.Open();
+
+            return int.Parse(conn.ServerVersion.Split(".")[0]);
+        }
+
         public static bool DatabaseExists()
         {
             dynamic result = new QueryFactory(new MySqlConnection($"SERVER={Server};PORT={Port};USER={User};PASSWORD={Password};"), new MySqlCompiler())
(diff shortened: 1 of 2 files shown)


MapleServer2/Database/DatabaseManager.cs
```diff
@@ -56,6 +56,14 @@ static DatabaseManager()
 
         public static void RunQuery(string query) => new QueryFactory(new MySqlConnection(ConnectionString), new MySqlCompiler()).Statement(query);
 
+        public static int getVersion()
```
AlanMorel, line 59: Should be `GetVersion`

AlanMorel (changes requested):

MapleServer2/MapleServer.cs
```diff
@@ -13,13 +13,12 @@
...
 using NLog;
 
-
 namespace MapleServer2
 {
 
     public static class MapleServer
     {
-        const int MIN_MYSQL_VERSION = 9;
+        const int MIN_MYSQL_VERSION = 8;
```
darkvergus, line 21: make it private

AngeloTadeucci (approved):

darkvergus (approved):

AlanMorel (approved):

Issue #92, opened 2020-12-30
Project style


integralcoder95:
### Handlers
   - All must implement LogUnknownMode if they have multiple modes.
   - All modes must be separated into different methods.
   - All must be stateless.

### Packets
   - Only use one file per Send/Recv OP.
   - All mode must have their own method with an enum for the mode byte.

## GameDataParser
   - Don't import MapleServer2 into GameDataParser project.
   - Metadatas must be readonly or init-only. You can use record for that.
  
## Maple2Storage
   - Files that can be used by GameDataParser and MapleServer2 should live here. (e. g.: Metadata files)

angelotadeucci:
All these has been done.

Create methods for common actions like warping and removing mesos/merets
Format all files in the codebase
Remove all uses of var from the source
Implement database (WIP)
Set unlocked titles as array in Player, replace hard-coded list

AlanMorel:
Updated

Sparkymod:
GameDataParser maintainable implementation to handle for future scalability #149

AngeloTadeucci:
Done:

> Rewrite Networking + Sessions (Use PipeScheduler) #456 
> Rewrite logging system
> Move handshake packet code to Packet file #456

PR #456, opened 2021-11-02
Feat: Pipe Scheduler

AngeloTadeucci:
Implemented Pipe Scheduler. #92 

### Notable changes

All files in MaplePacketLib2

Files in MapleServer2/Network

Everything else is basically renaming


MapleServer2/Packets/BannerListPacket.cs
```diff
@@ -25,7 +25,7 @@ public static Packet SetBanner(List<Banner> banners)
                 }
                 pWriter.WriteUnicodeString("");
                 pWriter.WriteUnicodeString(banner.ImageUrl);
-                pWriter.WriteEnum(banner.Language);
+                pWriter.WriteInt((int) banner.Language);
```
Sparkymod, line 28: does `pWriter.Write(banner.Language);` works here? since it was an enum.

AlanMorel (approved):

MapleServer2/Packets/BannerListPacket.cs
AngeloTadeucci: yeah mb

PR #481, opened 2021-11-16
Clean up: Refactor field objects, stats, and field manager.

quotopian:
In preparation of combat, pathfinding, skills reworking, and more. Basically separates game logic + data classes.

## Additions
- Non-generic `IFieldObject`
  - When you want to use any field object.
- `IFieldActor`, `IFieldActor<T>`
  - Actors are characters (think game characters that are movable, have stats, animations, etc..)
  - Think of the Actor as a game object, and `T` as the data to describe the actor (player, npc, mob...).
- `INpc`
  - Defines objects that are controllable by the AI (think CPU players). Couldn't find a better name.
  - FieldManager npcs and mobs implement `INpc`
- FieldManager's `FieldActor<T>`, `Character`, `Npc`, `Mob`
  - Classes that the FieldManager uses to be a factory for these objects (manage ObjectIds).
  - `Character` defines a `FieldActor<Player>`.  A Player's Character!
  - `Npc` defines an NPC character, but also not a mob.
  - `Mob` defines a Mob who runs an MobAI state machine to determine its actions. Also contains extra combat logic.
- Load player stats from database.
- Basic mob aggro logic.

## Removals
- `Npc`
  - This was a weird mix of metadata + coordinates that always had to be synced with the field object.
  - Just separate game logic into `IFieldActor` and an Npc's information in `NpcMetadata` to make `IFieldActor<NpcMetadata>`.
- `Mob`
  - Same as above.
- `ItemAttributes`
  - It was the same as `PlayerStat` (now `StatId`).

## Changes
- FieldManager's RequestFieldObject(), RequestCharacter(), RequestNpc(), RequestMob()
  - These methods are meant to be public and create their respective classes, possibly binding them to the field.
- FieldManager's AddPlayer(), AddNpc(), AddMob()
  - These methods are meant to bind the respective FieldObjects/FieldActors to this field.
  - Similar to how Remove() unbinds the object from the field.
- `PlayerStats` -> `Stats`
  - Defines stats for all actors or characters in the field (players, npcs, mobs...)
  - All stats are longs. Since it's common, `Bonus`, `Base`, `Total` refer to int representations. `BonusLong`, `BaseLong`, `BaseTotal` give the longs.
  - Easily pick a stat and increase/decrease it. (ex. `Stats[StatId.Hp].IncreaseBase(5)`)
- `DamageHandler`
  - Represents "effects" or the damage from a skill which can then be applied to an Actor to calculate damage using source's + target's stats.
## Known Bugs
- FieldPlayer's coord is not changed from previous field when using Warp and similar methods to a new map for the first time.
- Changing channels ultimately assigns a Player's character to the new instance, but the previous sessions disconnects and resets ObjectId to -1.
  - Possible fix: Wait until character is disconnected from previous map (ObjectId == -1), before session initialization.

**tl;dr**: I separated classes that act like data (Player.cs) and classes that are actually game objects in the world (IFieldObjects, IFieldActors), and generalized stats, skills, and damage a bit. This way we can add code to handle skills, damage, positions, movement, pathfinding for player characters and NPCs like mobs only inside the field code.


AngeloTadeucci:
You should merge upstream so we don't see the diff from commit <https://github.com/AlanMorel/MapleServer2/pull/481/commits/bf72125c9e057378748839bc9269f5dae7f67e5e>

MapleServer2/Types/Stats.cs
```diff
@@ -0,0 +1,545 @@
+﻿using Maple2Storage.Types;
+using Maple2Storage.Types.Metadata;
+
+namespace MapleServer2.Types;
+
+// Player Stats in Packet Order - Count: 35 (0x23)
+public enum StatId : byte
```
AngeloTadeucci, line 7: Delete this and use Maple2Storage.Enums.ItemAttribute? maybe rename the enum too

MapleServer2/Types/Stats.cs
```diff
@@ -0,0 +1,545 @@
...
+ * Change PlayerStat.Current for temporary changes.
+ */
+public class Stat
+{
+    public long BonusLong;
+    public long BaseLong;
+    public long TotalLong;
+    public int Bonus { get => (int) BonusLong; protected set { BonusLong = value; } }
+    public int Base { get => (int) BaseLong; protected set { BaseLong = value; } }
+    public int Total { get => (int) TotalLong; protected set { TotalLong = value; } }
```
AngeloTadeucci, line 417: Why this is needed when everything is long? Same for the methods below

AngeloTadeucci (changes requested):

MapleServer2/Types/Stats.cs
quotopian: Doing this allows us to use a stat in 1 type for dictionaries and as a int or as a long without type casting. We can't use class inheritance because it won't serialize.

AngeloTadeucci (approved):

darkvergus (approved):

MapleServer2/Types/Stats.cs
```diff
@@ -0,0 +1,506 @@
...
+    }
+}
+
+/* Max = Base + stat allocations + bonuses.
+ * Min = Base stat amount.
+ * Current = Final value (e.g. capped Damage, current Hp, active CritRate, ...)
+ *
+ * Change PlayerStat.Current for temporary changes.
+ */
+public class Stat
```
AlanMorel, line 371: Can this class be put into a separate file?

MapleServer2/Managers/FieldManager.cs
```diff
@@ -353,9 +348,70 @@ public IFieldObject<T> RequestFieldObject<T>(T wrappingObject)
...
+        mob.Rotation = rotation;
+        if (animation != default)
+        {
+            mob.Animation = animation;
+        }
+        AddMob(mob);
+        return mob;
+    }
+
+    public IFieldActor<NpcMetadata> RequestMob(int mobId, CoordF coord = default, CoordF rotation = default, short animation = default)
```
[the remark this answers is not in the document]
AlanMorel: If so, one method should just call the other

MapleServer2/Managers/FieldManager.cs
```diff
@@ -711,26 +793,608 @@ public FieldObject(int objectId, T value)
         }
     }
 
+    private abstract class FieldActor<T> : FieldObject<T>, IFieldActor<T>
```
AlanMorel, line 796: This file is already 700+ lines long, can this class be moved to a new file?

MapleServer2/Managers/FieldManager.cs
```diff
@@ -711,26 +793,608 @@ public FieldObject(int objectId, T value)
...
+            }
+        }
+
+        public virtual void Perish()
+        {
+            IsDead = true;
+        }
+    }
+
+    private class Character : FieldActor<Player>
```
AlanMorel, line 948: Actually, can a folder of actors just be made, and every private class thrown into a separate class inside that folder?

AlanMorel (changes requested):

darkvergus (approved):

AlanMorel (approved):

Sparkymod (approved):

PR #496, opened 2021-12-16
Fix: dynamic trophy update

AngeloTadeucci:
- Dynamic trophy update:
    - Find all trophies that use the condition type, code and target;
    - Update all trophies which match the conditions.
- Added a Level Up Command.

##
Worried about performance of the function **TrophyUpdate** because of searching trough all trophy metadatas to find the correct condition, I did some tests with this snippet:

```cs
Stopwatch sw = Stopwatch.StartNew();
for (int i = 0; i < 100; i++)
{
    trigger.Session.Player.TrophyUpdate("jump", 1);
}
sw.Stop();
```

The result was ~220ms for all 100 iterations with saving to database.

Without saving to database the time was ~8ms for 100 iterations.


darkvergus (approved):

MapleServer2/Commands/Game/PlayerCommands.cs
```diff
@@ -145,9 +153,31 @@ public override void Execute(GameCommandTrigger trigger)
...
+{
+    public LevelUpCommand()
+    {
+        Aliases = new()
+        {
+            "levelup"
+        };
+        Description = "Level up.";
+        Parameters = new()
+            ;
```
AlanMorel, line 171: Weird formatting?

AlanMorel (changes requested):

AlanMorel (approved):

Zintixx (approved):

PR #502, opened 2021-12-23
Cleanup

darkvergus:
- Removed redundant code that the compiler auto generates.
- Fixed lambda_methods.
- General Cleanup

### DIFF
--- a/MapleServer2/Enums/SystemNotice.cs
+++ b/MapleServer2/Enums/SystemNotice.cs
@@ -1,6 +1,6 @@
 ﻿namespace MapleServer2.Enums;
 
-public enum SystemNotice : int
+public enum SystemNotice
 {
     ChooseYourName = 0x1,
     Enter2Letters = 0xB,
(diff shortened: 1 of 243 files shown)


AngeloTadeucci (approved):

MapleServer2/Enums/SystemNotice.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace MapleServer2.Enums;
 
-public enum SystemNotice : int
+public enum SystemNotice
```
Zintixx, line 3: Like this and many others, has this PR even been tested? Without indicating it's an int, it's going to ruin a lot of packets being written. How will the server know when to write an int or not?

Zintixx (changes requested):

MapleServer2/Enums/SystemNotice.cs
darkvergus: by default its an int

AngeloTadeucci: Yes this has been tested and Rider says this: 'int' is default enum governing type

darkvergus: the compiler assumes its an int by default thats why its ok to remove it

Zintixx (approved):

AlanMorel (approved):

PR #504, opened 2021-12-27
Implement sending of guild mail

JamieDavidson:
Adds the ability to send guild mail
- Sends to all members including the sender
- Checks that the user has permission to do so, and displays appropriate error message if not
- Checks that the user is in a guild before sending, to avoid sneaky packet manipulating players
- Adds quick helper method to `GuildRank` to see if player's rank has appropriate flag set
  - This is useful in a couple other places in `GuildHandler`, but didn't want to create noise in this PR

### DIFF
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -555,6 +555,15 @@ public class GuildHandler : GamePacketHandler
         Player sender = session.Player;
         Guild guild = session.Player.Guild;
 
+        byte senderRank = sender.GuildMember.Rank;
+        GuildRank guildRank = guild.Ranks[senderRank];
+
+        if (!guildRank.HasRight(GuildRights.CanGuildMail))
+        {
+            session.Send(GuildPacket.ErrorNotice((byte)GuildErrorNotice.InsufficientPermissions));
+            return;
+        }
+
         IEnumerable<long> guildMemberCharacterIds = guild.Members
             .Select(m => m.Player.CharacterId)
             .Where(i => i != sender.CharacterId);
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -568,7 +568,7 @@ public class GuildHandler : GamePacketHandler
             session.Send(GuildPacket.ErrorNotice((byte) GuildErrorNotice.InsufficientPermissions));
             return;
         }
-        
+
         session.Send(GuildPacket.SendMail());
 
         IEnumerable<long> recipientIds = guild.Members.Select(c => c.Player.CharacterId);
--- a/MapleServer2/PacketHandlers/Game/GuildHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/GuildHandler.cs
@@ -4,8 +4,6 @@ using MapleServer2.Constants;
 using MapleServer2.Data.Static;
 using MapleServer2.Database;
 using MapleServer2.Enums;
-using MapleServer2.Managers;
-using MapleServer2.Network;
 using MapleServer2.PacketHandlers.Game.Helpers;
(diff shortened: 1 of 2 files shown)


MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -541,6 +547,35 @@ private static void HandleListGuild(GameSession session, PacketReader packet)
         session.Send(GuildPacket.ListGuildUpdate(session.Player, toggle));
     }
 
+    private static void HandleGuildMail(GameSession session, IPacketReader packet)
+    {
+        string title = packet.ReadUnicodeString();
+        string body = packet.ReadUnicodeString();
+
+        Player sender = session.Player;
+        Guild guild = session.Player.Guild;
```
AngeloTadeucci, line 556: I would check for null since we can't trust the client also some kind of anti packet editing.

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -541,6 +547,35 @@ private static void HandleListGuild(GameSession session, PacketReader packet)
...
+
+        IEnumerable<long> guildMemberCharacterIds = guild.Members
+            .Select(m => m.Player.CharacterId)
+            .Where(i => i != sender.CharacterId);
+        
+        foreach (long characterId in guildMemberCharacterIds)
+        {
+            MailHelper.SendMail(MailType.Player, characterId, sender.CharacterId, sender.Name, title, body, "", "", new(), 0, 0, out Mail mail);
+
+            session.Send(MailPacket.Send(mail));
```
AngeloTadeucci, line 575: it's meant to send the same packet several times to the sender?

AngeloTadeucci (changes requested):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
JamieDavidson: Done, good spot :)

MapleServer2/PacketHandlers/Game/GuildHandler.cs
```diff
@@ -541,6 +547,40 @@ private static void HandleListGuild(GameSession session, PacketReader packet)
         session.Send(GuildPacket.ListGuildUpdate(session.Player, toggle));
     }
 
+    private static void HandleGuildMail(GameSession session, IPacketReader packet)
+    {
+        string title = packet.ReadUnicodeString();
+        string body = packet.ReadUnicodeString();
+
+        Player sender = session.Player;
+        Guild guild = session.Player.Guild;
```
AngeloTadeucci, line 556:
```suggestion
        Guild guild = sender.Guild;
```

MapleServer2/PacketHandlers/Game/GuildHandler.cs
JamieDavidson: Maybe I've misunderstood the point of the MailPacket? From the wording, I understand it to be some sort of confirmation that the client is waiting for to confirm that the mail request finished being processed.

Is it perhaps the packet that drives the "You've got mail" notification that pops up? If this is the case, I believe there's a bug in the regular MailHandler too, I admit this was a lazy copy paste from [MailHandler::HandleSend:122](https://github.com/AlanMorel/MapleServer2/blob/master/MapleServer2/PacketHandlers/Game/MailHandler.cs#L122)

AngeloTadeucci: I think you should ask for how it's supposed to work in discord since we can test it in the official server and also sniff to see what exactly the server sends.

darkvergus (approved):

MapleServer2/PacketHandlers/Game/GuildHandler.cs
JamieDavidson: This should be good to go now

Sparkymod (approved):

AngeloTadeucci (approved):

PR #506, opened 2021-12-27
Fix: check for trophy range conditions

AngeloTadeucci:
- Handling type "map" for trophies;
- Function IsInConditionRange made by @JamieDavidson

Co-Authored-By: JamieDavidson <JamieDavidson@users.noreply.github.com>

### DIFF
--- a/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
+++ b/MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
@@ -23,13 +23,14 @@ public class FieldEnterHandler : GamePacketHandler
         Player player = session.Player;
         Account account = player.Account;
         session.EnterField(player);
-        session.Send(StatPacket.SetStats(session.Player.FieldPlayer));
+        session.Send(StatPacket.SetStats(player.FieldPlayer));
         session.Send(StatPointPacket.WriteTotalStatPoints(player));
 
         if (account.IsVip())
         {
-            session.Send(BuffPacket.SendBuff(0, new(100000014, session.Player.FieldPlayer.ObjectId, session.Player.FieldPlayer.ObjectId, 1, (int) account.VIPExpiration, 1)));
-            session.Send(PremiumClubPacket.ActivatePremium(session.Player.FieldPlayer, account.VIPExpiration));
+            session.Send(BuffPacket.SendBuff(0,
+                new(100000014, player.FieldPlayer.ObjectId, player.FieldPlayer.ObjectId, 1, (int) account.VIPExpiration, 1)));
+            session.Send(PremiumClubPacket.ActivatePremium(player.FieldPlayer, account.VIPExpiration));
         }
 
         session.Send(EmotePacket.LoadEmotes(player));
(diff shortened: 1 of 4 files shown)


darkvergus (approved):

AlanMorel (approved):

MapleServer2/Types/Trophy.cs
```diff
@@ -134,4 +134,25 @@ private static bool RewardTypeRequiresClaim(RewardType type)
             _ => false
         };
     }
+
+    public static bool IsInConditionRange(string trophyCondition, string condition)
```
JamieDavidson, line 138: This method doesn't feel like it should live here IMHO. If it were testing against the property `GradeCondition` I'd say put it here. To me, it feels like it should live in either `TrophyMetaDataStorage` or some new class that wraps these trophies.

AngeloTadeucci: moved it

darkvergus (approved):

MapleServer2/Types/Player.cs
```diff
@@ -432,7 +432,7 @@ public void TrophyUpdate(string type, int addAmount, string code = "", string ta
 
             if (trophy.GradeCondition.ConditionCodes[0].Contains('-'))
```
JamieDavidson, line 433: Not one for this PR, but something to consider - this will need to be expanded upon to work for some trophies, such as some item collection trophies that require you to (for example) loot all of X items. These include multiple condition codes, one for each item.

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -56,5 +59,9 @@ public override void Handle(GameSession session, PacketReader packet)
 
         List<GameEvent> gameEvents = DatabaseManager.Events.FindAll();
         session.Send(GameEventPacket.Load(gameEvents));
+
+        // get map id as string and size of 8 digits
+        string mapId = player.MapId.ToString().PadLeft(8, '0');
+        player.TrophyUpdate(type: "map", addAmount: 1, code: mapId);
```
JamieDavidson, line 65: You might have already seen this - but the "map" category is actually pretty small, I get the impression that it was used early on and then abandoned in favour of the "explore" category. Not something to worry about for this PR though :)

MapleServer2/PacketHandlers/Game/FieldEnterHandler.cs
```diff
@@ -23,31 +23,34 @@ public override void Handle(GameSession session, PacketReader packet)
...
-        session.Send(StatPacket.SetStats(session.Player.FieldPlayer));
+        session.Send(StatPacket.SetStats(player.FieldPlayer));
         session.Send(StatPointPacket.WriteTotalStatPoints(player));
 
         if (account.IsVip())
         {
-            session.Send(BuffPacket.SendBuff(0, new(100000014, session.Player.FieldPlayer.ObjectId, session.Player.FieldPlayer.ObjectId, 1, (int) account.VIPExpiration, 1)));
-            session.Send(PremiumClubPacket.ActivatePremium(session.Player.FieldPlayer, account.VIPExpiration));
+            session.Send(BuffPacket.SendBuff(0,
+                new(100000014, player.FieldPlayer.ObjectId, player.FieldPlayer.ObjectId, 1, (int) account.VIPExpiration, 1)));
```
Zintixx, line 32: Maybe doesn't have to be in this PR but the buff isn't 100000014 in GMS. It's 100000043 and 100000045. These should be parsed from /table/vipbenefiteffecttable.xml

AngeloTadeucci: Lets do it in another PR!

PR #510, opened 2021-12-30
Implement new trophy system, and many different trophy categories

JamieDavidson:
This felt like it didn't belong in the player class, and made a few assumptions about trophy conditions.
This is technically WIP, but feels ok to PR now as it's a replacement for an existing system that wasn't finished.

Introduces `TrophyManager`
- Event based, based on packet handler events
- Will likely have one method per category eventually
- Handles conditions that are a single number, a range of numbers, or comma separated lists of numbers. These are the 3 types we've come across so far.

Does not handle trophies that require progress tracking, e.g `ItemCollect` trophies.

PR #517, opened 2022-01-19
Fix: Trophy Manager

AngeloTadeucci:
- Handling trophy targets which was missing from #510


Zintixx (approved):

Sparkymod (approved):

MapleServer2/Managers/TrophyManager.cs
```diff
@@ -42,13 +42,13 @@ public static void OnPlayTimeTick(Player player)
...
     public static void OnLevelUp(Player player)
     {
-        int jobId = (int) player.JobCode;
+        int jobId = (int) player.Job;
+        int level = player.Levels.Level;
         IEnumerable<TrophyMetadata> levelUpTrophies = GetRelevantTrophies(TrophyTypes.LevelUp);
         IEnumerable<TrophyMetadata> levelTrophies = GetRelevantTrophies(TrophyTypes.Level);
 
-        // TODO: This might not be right :D
         IEnumerable<TrophyMetadata> matchingTrophies = levelUpTrophies
```
JamieDavidson, line 50: Now that I look back on this, I don't think I named these collections very well :)

Could we maybe rename `levelUpTrophies` -> `jobSpecificLevelTrophies` and `levelTrophies` to `generalLevelTrophies`? We could probably also merge the call to GetRelevantTrophies and the `.Where` filter, but this is all just personal preference :D

MapleServer2/Managers/TrophyManager.cs
```diff
@@ -42,13 +42,13 @@ public static void OnPlayTimeTick(Player player)
...
-            .Where(t => IsMatchingCondition(t.Grades.First().ConditionCodes, jobId));
+            .Where(t => IsMatching(t.Grades.First().ConditionCodes, jobId) && IsMatching(t.Grades.First().ConditionTargets, level));
```
JamieDavidson, line 51: Have we checked the trophies for the progress steps to see that they only have 1? I expect "Reach X level on Y job" to have only 1 progress step, so this will only add 1 to progress upon hitting the right level, but who knows with how crazy this game is sometimes 🤷🏻‍♂️

MapleServer2/Managers/TrophyManager.cs
```diff
@@ -85,66 +85,55 @@ public static void OnTrigger(Player player, string trigger)
...
 
-    private static bool IsMatchingCondition(IEnumerable<string> trophyConditions, string condition)
+    private static IEnumerable<TrophyMetadata> GetRelevantTrophies(string category) => TrophyMetadataStorage.GetTrophiesByType(category);
+
+    /// <summary>
+    /// Checks if the given condition code or target match the given string value.
+    /// </summary>
+    /// <param name="trophyCondition">Trophy code or target</param>
+    /// <param name="value">Value to try match</param>
+    /// <returns>True or false</returns>
```
JamieDavidson, line 102: Big ❤️ for including XML documentation, anything that helps intellisense is a win in my book.
I think it's definitely worth having a talk about what level of detail we want from these. For example, IMHO "Returns true or false" is obvious enough by nature of the method returning a boolean, and so I wouldn't have added the `<returns>` tag myself. Thank you for taking charge and starting positive change 😉

MapleServer2/Types/Trophy.cs
```diff
@@ -65,7 +64,7 @@ public Trophy(long uid, int trophyId, int nextGrade, long counter, bool isDone,
         AccountId = accountId;
     }
 
-    public void AddCounter(GameSession session, long amount)
+    public void AddCounter(Player player, long amount)
```
JamieDavidson, line 67: This is a great example of passing around the minimum required class to perform an operation!

I see this a lot in server emulators in general - because people _can_ pass around an entire Session (or equivalent) object they will, when in reality the consumer only needs a small part of that object.

We have a few examples around the server where we only need an inventory, but we pass in an entire player object, and the first thing we do is get the inventory from the player!

Great job seeing this and improving it :)

JamieDavidson (approved):
🎉 
Great job :) couple questions for you, couple bits to think over, but looks good to me!

AlanMorel (approved):

MapleServer2/Managers/TrophyManager.cs
AngeloTadeucci: I was thinking about that, i feel it would make it better to read also.

PR #525, opened 2022-01-31
Fix: Region skills

AngeloTadeucci:
- Fixed region skills that paint the ground, like priest's Sanctuary
    - Don't add the effect on liquid.
    - Match the height of the blocks correctly

### DIFF
--- a/MapleServer2/Data/Static/MapMetadataStorage.cs
+++ b/MapleServer2/Data/Static/MapMetadataStorage.cs
@@ -122,8 +122,8 @@ public static class MapMetadataStorage
             return 99;
         }
 
-        // checking 20 blocks in the same Z axis
-        for (int i = 0; i < 20; i++)
+        // checking 10 blocks in the same Z axis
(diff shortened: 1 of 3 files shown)


Zintixx (approved):

MapleServer2/Data/Static/MapMetadataStorage.cs
```diff
@@ -73,8 +89,52 @@ public static int GetPlotNumber(int mapId, CoordB coord)
...
+        CoordS tempCoord = coord;
+        MapMetadata mapMetadata = GetMetadata(mapId);
+        if (mapMetadata is null)
+        {
+            block = null;
+            return 99;
+        }
+
+        // checking 20 blocks in the same Z axis
+        for (int i = 0; i < 20; i++)
```
darkvergus, line 126: is there really a need to check 20 blocks considering the skills only go down up to 2?

darkvergus (approved):

PR #552, opened 2022-02-23
WIP: Attendance Event, Event User Values, and Events Rework

Zintixx:
- This implements (and activates) attendance event. I'm only saying WIP because a lot of things may change.
 - It is fully functional but however, the backend of things likely need to be shifted around, particularly Event User Values. With this being the first event using event user values as of now, things may be entirely reworked once more to get a better understanding on how they're used in a more global scale.
 - Moved over updating a player's log time out of the OnlineTimer loop in order for it to be of use for AttendanceEvent.
 - temporarily hardcoding a attendance mail. This likely needs something more flexible to support more custom mails


MapleServer2/Types/GameEventUserValue.cs
```diff
@@ -0,0 +1,34 @@
...
+using MapleServer2.Enums;
+
+namespace MapleServer2.Types;
+
+public class GameEventUserValue
+{
+    public long Id;
+    public long CharacterId;
+    public GameEventUserValueType EventType;
+    public string EventValue;
```
AngeloTadeucci, line 11: So this value is always a number but in the packet is always a string? If that's the case I would make this a int and only use ToString() on the packet.

MapleServer2/Types/TimeInfo.cs
```diff
@@ -7,4 +7,8 @@ public static class TimeInfo
     public static DateTime CurrentDate() => DateTimeOffset.UtcNow.UtcDateTime;
 
     public static long AddDays(int days) => DateTimeOffset.UtcNow.AddDays(days).ToUnixTimeSeconds();
+
+    public static long Tomorrow() => (long) DateTime.Now.AddDays(1).ToUniversalTime().Date.Subtract(new DateTime(1970, 1, 1)).TotalSeconds;
+
+    public static DateTime TimestampToDateTime(long timestamp) => new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc).AddSeconds(timestamp);
```
AngeloTadeucci, line 13: Can't you just parse the timestamp???

MapleServer2/Types/TimeInfo.cs
```diff
@@ -7,4 +7,8 @@ public static class TimeInfo
     public static DateTime CurrentDate() => DateTimeOffset.UtcNow.UtcDateTime;
 
     public static long AddDays(int days) => DateTimeOffset.UtcNow.AddDays(days).ToUnixTimeSeconds();
+
+    public static long Tomorrow() => (long) DateTime.Now.AddDays(1).ToUniversalTime().Date.Subtract(new DateTime(1970, 1, 1)).TotalSeconds;
```
AngeloTadeucci, line 11: Why substract 0?

AngeloTadeucci (changes requested):

MapleServer2/Types/TimeInfo.cs
AngeloTadeucci:
```csharp
        ((DateTimeOffset) DateTime.Now.AddDays(1).Date).ToUnixTimeSeconds(); // without timezone, without time
        ((DateTimeOffset) DateTime.Now.AddDays(1).ToUniversalTime().Date).ToUnixTimeSeconds(); // UTC, without time
        DateTimeOffset.UtcNow.AddDays(1).ToUnixTimeSeconds(); // utc, with time
```

MapleServer2/Types/TimeInfo.cs
AngeloTadeucci: `DateTimeOffset.FromUnixTimeSeconds(unixTimestamp)`

MapleServer2/Types/GameEventUserValue.cs
Zintixx: not always an int. It's sometimes a bool, int, or long

AngeloTadeucci (approved):

PR #561, opened 2022-02-28
Extract Inventory interface - Make items private, restrict access to methods only

JamieDavidson:
Extracts interface for player inventory

Makes Items collection private, we should not allow the entire codebase to do whatever it likes with the underlying collection, as this makes it easy to introduce bugs.

Introduces some common methods to replace cases where other classes were "dipping in to" the underlying collection. This has some benefits:

E.g. Reduces cognitive complexity when reading code at a glance
```csharp
Item itemCost = session.Player.Inventory.Items.FirstOrDefault(x => x.Value.Id == requiredItemId).Value;
```
Vs
```csharp
Item itemCost = session.Player.Inventory.GetById(requiredItemId);
```
Removes Dictionary-specific behaviour outside of Inventory class, now the underlying collection can easily be replace with e.g. a List if required:
```csharp
if (!inventory.Items.TryGetValue(itemUid, out Item item))
{
    return;
}
// GMS2 Always 100% success rate
session.Send(LapenshardPacket.Select(10000));
```
vs
```csharp
if (!inventory.HasItem(itemUid))
{
    return;
}
// GMS2 Always 100% success rate
session.Send(LapenshardPacket.Select(10000));
```

Has potential for introducing more methods, e.g.
```csharp
Player.Inventory.GetAllById(itemId)
            .Sum(x => x.Amount);
```
Could be replaced with
```csharp
Player.Inventory.GetItemAmount(itemId);
```

### DIFF
--- a/MapleServer2/Types/IInventory.cs
+++ b/MapleServer2/Types/IInventory.cs
@@ -10,6 +10,7 @@ public interface IInventory
     bool RemoveItem(GameSession session, long uid, out Item item);
     void DropItem(GameSession session, long uid, int amount, bool isBound);
     void MoveItem(GameSession session, long uid, short dstSlot);
+    bool HasItem(long uid);
     bool Replace(Item item);
     void SortInventory(GameSession session, InventoryTab tab);
     void LoadInventoryTab(GameSession session, InventoryTab tab);
--- a/MapleServer2/Types/IInventory.cs
+++ b/MapleServer2/Types/IInventory.cs
@@ -5,13 +5,27 @@ namespace MapleServer2.Types;
 
 public interface IInventory
 {
+    Dictionary<InventoryTab, short> ExtraSize { get; }
+    long Id { get; }
+    Dictionary<ItemSlot, Item> Equips { get; }
+    Dictionary<ItemSlot, Item> Cosmetics { get; }
+    Item[] Badges { get; }
+    Item[] LapenshardStorage { get; }
+    Dictionary<long, Item> TemporaryStorage { get; }
+    
     void AddItem(GameSession session, Item item, bool isNew);
     void ConsumeItem(GameSession session, long uid, int amount);
     bool RemoveItem(GameSession session, long uid, out Item item);
     void DropItem(GameSession session, long uid, int amount, bool isBound);
     void MoveItem(GameSession session, long uid, short dstSlot);
     bool HasItem(long uid);
+    bool HasItem(int id);
     Item GetByUid(long uid);
+    Item GetById(int id);
+    IReadOnlyCollection<Item> GetItemsNotNull();
+    IReadOnlyCollection<Item> GetAllById(int id);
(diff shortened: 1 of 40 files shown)


MapleServer2/Types/IInventory.cs
```diff
@@ -0,0 +1,38 @@
...
+
+public interface IInventory
+{
+    Dictionary<InventoryTab, short> ExtraSize { get; }
+    long Id { get; }
+    Dictionary<ItemSlot, Item> Equips { get; }
+    Dictionary<ItemSlot, Item> Cosmetics { get; }
+    Item[] Badges { get; }
+    Item[] LapenshardStorage { get; }
+    Dictionary<long, Item> TemporaryStorage { get; }
```
AngeloTadeucci, line 14: I think if you remove all of these except Id, we could use this interface for the others inventories. Maybe in the future, not now.

MapleServer2/Types/Inventory.cs
```diff
@@ -323,10 +323,52 @@ public void MoveItem(GameSession session, long uid, short dstSlot)
         session.Send(ItemInventoryPacket.Move(dstUid, srcSlot, uid, dstSlot));
     }
 
+    public bool HasItem(long uid)
```
AngeloTadeucci, line 325: Can we use expression body for all these functions?

AngeloTadeucci (changes requested):

MapleServer2/Types/IInventory.cs
JamieDavidson: That's the plan! I wanted to keep this PR succinct, and then have other PRs to handle these.
IMHO they should be separated out, ie Inventory should not handle Equipment, but that's a discussion for later :P

MapleServer2/Types/Inventory.cs
JamieDavidson: Sure can!

AngeloTadeucci (approved):

darkvergus (approved):

MapleServer2/Types/IInventory.cs
```diff
@@ -0,0 +1,81 @@
...
+    void MoveItem(GameSession session, long uid, short dstSlot);
+
+    /// <summary>
+    /// Determines whether the inventory contains an item with given Unique ID (UID)
+    /// </summary>
+    /// <param name="uid">The UID of the item</param>
+    bool HasItem(long uid);
+
+    /// <summary>
+    /// Determins whether the inventory contains an item with given Item ID
```
AngeloTadeucci, line 28:
```suggestion
    /// Determines whether the inventory contains an item with given Item ID
```

AngeloTadeucci (changes requested):

AngeloTadeucci (approved):

Zintixx (approved):

Sparkymod (approved):

PR #566, opened 2022-03-04
Triggers

AngeloTadeucci:
- #508 

## Main changes
- Implemented **FaceEmotion**;
- Implemented **CreateMonster** with spawn animation;
- Fixed conversation trigger;
- Using the correct exception for .env file not found.


darkvergus (approved):

Zintixx (approved):

MapleServer2/Packets/CinematicPacket.cs
```diff
@@ -60,12 +60,12 @@ public static PacketWriter StartSceneSkip()
         return pWriter;
     }
 
-    public static PacketWriter Conversation(int npcId, string stringId, int delay, Align align)
+    public static PacketWriter Conversation(int npcId, string illustId, string stringId, int delay, Align align)
```
AlanMorel, line 63: Minor but does `illust` mean `illustration`? If so, probably better to name it that, if not then nevermind

AlanMorel (changes requested):

MapleServer2/Packets/CinematicPacket.cs
AngeloTadeucci: It is, i just copied from the trigger and didn't think about it. Will fix tomorrow and merge

PR #567, opened 2022-03-05
Scheduled Global Events

Zintixx:
- Implementing scheduled global events. This follows what the MS2 wiki has documented.
- GlobalEventCommand now lets you use 1 or 2 event IDs instead of needing to need 3
- Moving player out of an instanced and non tutorial map if they log out in said map.
- Removed some error packets for attendance (client already sends it)


AngeloTadeucci (approved):

MapleServer2/Data/Static/MapMetadataStorage.cs
```diff
@@ -138,4 +138,26 @@ public static int GetDistanceToNextBlockBelow(int mapId, CoordS coord, out MapBl
...
+    public static bool MapIsTutorial(int mapId)
+    {
+        MapMetadata mapMetadata = GetMetadata(mapId);
+        if (mapMetadata is null)
+        {
+            return false;
+        }
+
+        return mapMetadata.Property.IsTutorialMap;
+    }
```
JamieDavidson, line 162: IMO this should have been a separate PR, as it doesn't relate to event changes. Up to you if you want to split it out.

MapleServer2/Managers/GlobalEventManager.cs
```diff
@@ -30,4 +32,19 @@ public GlobalEvent GetEventById(int id)
     {
         return GlobalEventList.GetValueOrDefault(id);
     }
+
+    public GlobalEvent GetCurrentEvent()
+    {
+        return GlobalEventList.Values.FirstOrDefault();
```
JamieDavidson, line 38: It doesn't look like the current event is guaranteed to be the first one to me, if someone schedules an event which is going to run in 10 hours, then schedules another to run in 1 hour, the second will run first.

MapleServer2/MapleServer.cs
```diff
@@ -54,7 +54,7 @@ public static async Task Main()
         }
 
         // Schedule daily reset and repeat every 24 hours
-        TaskScheduler.Instance.ScheduleTask(0, 0, 24, DailyReset);
+        TaskScheduler.Instance.ScheduleTask(0, 0, 24, 24 * 60, DailyReset);
```
JamieDavidson, line 57: Have you considered swapping these parameter types out for `TimeSpan`s? I find the parameter list quite confusing - we have a timestamp in its separate parts, then a parameter in minutes. Not sure if this would make it better or worse - just a suggestion.

MapleServer2/Tools/TaskScheduler.cs
```diff
@@ -9,13 +9,13 @@ public class TaskScheduler
...
-    public void ScheduleTask(int hour, int min, double intervalInHour, Action task)
+    public void ScheduleTask(int hour, int min, double intervalInHour, int minutesToRunPerDay, Action task)
     {
         DateTime now = DateTime.Now;
         DateTime firstRun = new(now.Year, now.Month, now.Day, hour, min, 0, 0);
-        if (now > firstRun)
+        while (now > firstRun)
         {
-            firstRun = firstRun.AddDays(1);
+            firstRun = firstRun.AddMinutes(minutesToRunPerDay);
```
JamieDavidson, line 18: I'm not sure I follow the logic here, I think this was fine before?
We schedule an event for a certain time during the day.
If that time has already passed today, we add 1 day on to the event.
We then figure out how long it's going to be until we run it next, and tell the timer to run for the first time at that time, and then at the interval after.

MapleServer2/Types/GlobalEvent.cs
```diff
@@ -1,42 +1,60 @@
...
     }
 
-    public async Task Start()
+    public void Start()
     {
-        MapleServer.BroadcastPacketAll(GlobalPortalPacket.Notice(this));
+        // Assign new ID
+        Id = GuidGenerator.Int();
+        GameServer.GlobalEventManager.AddEvent(this);
+        BroadcastEvent();
```
JamieDavidson, line 26: You have asynchronous code here that is not being awaited, which may cause various issues in a multi threaded application like this one. Start ought to be asynchronous.

MapleServer2/Types/GlobalEvent.cs
```diff
@@ -1,42 +1,60 @@
...
+    s_massive_event_name_crazy_runner = 4,
+    s_massive_event_name_final_surviver = 5,
+    s_massive_event_name_great_escape = 6,
+    s_massive_event_name_dancedance_stop = 7,
+    s_massive_event_name_crazy_runner_shanghai = 8,
+    s_massive_event_name_hideandseek = 9,
+    s_massive_event_name_red_arena = 10,
+    s_massive_event_name_blood_mine = 11,
+    s_massive_event_name_treasure_island = 12,
+    s_massive_event_name_christmas_dancedance_stop = 13
```
JamieDavidson, line 59: IMO this was better before - the Enum values are harder to read now, and harder to find the value you want via intellisense.

JamieDavidson (changes requested):
I think we need to revisit some of the logic here :D

JamieDavidson:
Also I couldn't add a comment to this line because it wasn't changed in this PR, but if the intention is that people no longer need to supply 3 events with the EventCommand (which I don't think was the case to begin with) then the help text is now misleading. It says "No events provided. Choose 3 events."

MapleServer2/Commands/Game/EventCommands.cs
```diff
@@ -54,8 +53,18 @@ public override void Execute(GameCommandTrigger trigger)
...
+        if (globalEvent.Events.Count == 1)
+        {
+            globalEvent.Events.Insert(0, GlobalEventType.none);
+            globalEvent.Events.Insert(2, GlobalEventType.none);
+
+        }
+        else if (globalEvent.Events.Count == 2)
+        {
+            globalEvent.Events.Insert(2, GlobalEventType.none);
+        }
```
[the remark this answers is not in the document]
Zintixx: It's working with Nexlul coding.
The event window needs to have 3 events passed through. However, there are occasions that the window does not use all 3 buttons, and this is a solution I figured would work.
When there is only one event passed through, it must use the middle slot (index 1).
If it uses two events, it uses the top 2 slots (index 0, and 1).
Example using 1 event:
image

Example of a standard 3 event:
image

Yes, you can technically try and register an event with more than 3, but it wont display if it's more than 3.

Maple2Storage/Json/GlobalEvents.json
```diff
@@ -0,0 +1,46 @@
+[
+  {
+    "Id": 1,
+    "FirstHour": 0,
+    "FirstMinutesOnHour": 5,
+    "MinutesToRunPerDay": 90,
+    "Events": [
+      12,
+      6,
+      1
```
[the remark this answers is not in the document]
Zintixx: I agree using the enum type would be better, but I would like to not hard code this. This should be flexible for server users to modify in case they want to add or manipulate the standard events.

MapleServer2/Managers/GlobalEventManager.cs
Zintixx: That's the point, isn't it? You want to get the current event. 
A scheduled event doesn't add the global event into the event manager until the scheduled time triggers.

MapleServer2/Tools/TaskScheduler.cs
Zintixx: These events run multiple times per day. If I kept it as days, the first time the server starts, the events would never trigger on the first day.

MapleServer2/Types/GlobalEvent.cs
AngeloTadeucci: I think these names are from xml, I'm correct? @Zintixx

Zintixx: They are but I can revert to what it was before, to regain that readability (removing the "s_massive_event_name_" part of the name and doing something like this:
image

AngeloTadeucci: What if you revert the naming and just create a function that inputs the enum and outputs the string `"s_massive_event_name_" + eventType`?

AlanMorel (approved):

MapleServer2/Managers/GlobalEventManager.cs
JamieDavidson: My mistake, I hadn't seen that `EventCommand` no longer calls `AddEvent` immediately.

JamieDavidson (approved):

PR #569, opened 2022-03-06
Cleanup: QuestManager & TrophyManager

AngeloTadeucci:
Extracting helper functions to its own helper class.

### DIFF
--- a/MapleServer2/Tools/ConditionHelper.cs
+++ b/MapleServer2/Tools/ConditionHelper.cs
@@ -5,7 +5,7 @@ public static class ConditionHelper
     /// <summary>
     /// Checks if the given condition code or target match the given string value.
     /// </summary>
-    /// <param name="conditionType">Trophy code or target</param>
+    /// <param name="conditionType">Condition code or target</param>
     /// <param name="value">Value to try match</param>
     public static bool IsMatching(string conditionType, string value)
     {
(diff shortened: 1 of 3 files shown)


MapleServer2/Tools/ConditionHelper.cs
```diff
@@ -0,0 +1,84 @@
...
+        // Check if the value is in the list
+        if (conditionType.Contains(',') && IsInList(conditionType, value))
+        {
+            return true;
+        }
+
+        return long.TryParse(conditionType, out long parsedCondition) && parsedCondition == value;
+    }
+
+    public static bool IsInList(string conditionList, long value)
```
JamieDavidson, line 55: Do these need to be public? I think we could probably get away with only `IsMatching` being public?

AngeloTadeucci: just fixed

AngeloTadeucci: Wait sorry, I was thinking in making it private but decided it could be useful in the future.

JamieDavidson: I don't think those methods will ever be useful publicly. In order to figure out which one to use, you'll have to duplicate the logic of for example seeing if there's a `-` in them. I think by making those methods public, we'll be encouraging people to duplicate that logic.

AngeloTadeucci: We never know that the company has in stock for us but sure i'll make them private

JamieDavidson (approved):

PR #588, opened 2022-03-28
Feat: Calculated Item Stats and Gear Score

Zintixx:
- Implements the lua scripts to calculate item stats and gear score.
~- Currently does not calculate a static defense value correctly. Still need to figure out how this is actually calculated~
- Rework on Item Stats to support all 3 item stats (constants, statics, randoms)
  - Suggestions to implement this is happily accepted. I do not think my solution is the best but it works.
 - Stats now accept any stat attributes as opposed to just the defined ones


MapleServer2/Types/ItemStats/StaticStats.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        if (result.Tuple[0].Number == 0 && result.Tuple[1].Number == 0)
+        {
+            return 0;
+        }
+
+        // Get random between min and max values
+        double statValue = random.NextDouble() * (result.Tuple[1].Number - result.Tuple[0].Number) + result.Tuple[0].Number;
+        return statValue;
```
AngeloTadeucci, line 153:
```suggestion
        return random.NextDouble() * (result.Tuple[1].Number - result.Tuple[0].Number) + result.Tuple[0].Number;
```

AngeloTadeucci (approved):

AlanMorel (approved):

AlanMorel (approved):

AngeloTadeucci (approved):

darkvergus (approved):

PR #598, opened 2022-04-08
Feat: Limit Break

Zintixx:
- Implements limit break (sorry)
- Added a proper enchant restriction check from the xml tables
- Made use of ItemSocketMetadata/Storage.
- Created a copy stat function to be used when an item is copied (in order for to make a new ItemStat class)


AlanMorel (approved):

GameDataParser/Parsers/EnchantLimitParser.cs
```diff
@@ -0,0 +1,50 @@
...
+    protected override List<EnchantLimitMetadata> Parse()
+    {
+        List<EnchantLimitMetadata> items = new();
+
+        foreach (PackFileEntry entry in Resources.XmlReader.Files)
+        {
+            if (!entry.Name.StartsWith("table/enchantlimittable"))
+            {
+                continue;
+            }
```
AngeloTadeucci, line 23: If its only one file that needs parsing, can we use a FirstOrDefault instead of ForEach?

MapleServer2/PacketHandlers/Game/LimitBreakHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    inventory.ConsumeItem(session, item.Uid, ingredient.Amount);
+                    break;
+                }
+
+                ingredient.Amount -= item.Amount;
+                inventory.ConsumeItem(session, item.Uid, item.Amount);
+            }
+        }
+    }
```
AngeloTadeucci, line 155: I think this is duplicate code right? We could move it to Inventory.

AngeloTadeucci (changes requested):

darkvergus (approved):

PR #604, opened 2022-04-16
Rework: FieldActors

AngeloTadeucci:
- Changed **FieldActors** to be public instead of `private partial`;
- Some minor logic changes to **FieldManager** like using LINQ to check for null objects.

### DIFF
--- a/MapleServer2/Managers/Actors/FieldActor.cs
+++ b/MapleServer2/Managers/Actors/FieldActor.cs
@@ -5,177 +5,174 @@ using MapleServer2.Packets;
 using MapleServer2.Servers.Game;
 using MapleServer2.Types;
 
-namespace MapleServer2.Managers;
+namespace MapleServer2.Managers.Actors;
 
-public partial class FieldManager
+public abstract class FieldActor<T> : FieldObject<T>, IFieldActor<T>, IDisposable
 {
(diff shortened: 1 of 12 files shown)


Sparkymod (approved):

Zintixx (approved):

MapleServer2/Managers/Actors/FieldActor.cs
```diff
@@ -5,177 +5,174 @@
...
+    {
+        if (amount <= 0)
         {
-            // TODO: Perform pathfinding, then set path.
-            Velocity = target - Coord;
+            return;
         }
 
-        public virtual void Cast(SkillCast skillCast)
+        lock (Stats)
```
AlanMorel, line 69: Why do we need locks?

AngeloTadeucci: good question, I believe is to ensure thread safety since Stats is a simple Dictionary

darkvergus (approved):

MapleServer2/Managers/Actors/FieldActor.cs
AlanMorel: aren't there concurrency-safe dictionaries in C#?

AngeloTadeucci: There are, I can change it in the next PR

PR #629, opened 2022-05-17
Rework: Stats packet

AngeloTadeucci:
- New: Add delay between mob dying and mob disappearing from field;
- Fix: Properly remove mob from field when dead;
- Fix: Remove velocity if mob is dead;
- Fix: NpcMetadata not all elements had **Order** in **[XmlElement]**.


darkvergus (approved):

MapleServer2/Managers/Actors/Mob.cs
```diff
@@ -149,7 +151,7 @@ private void Move(MobMovement moveType)
                     }
 
                     NextMovementTarget = coordF;
-                    Distance = (coordF - Coord.ToShort()).Length();
+                    Distance = (coordF.ToFloat() - Coord).Length();
```
Sparkymod, line 154: isn't this `coordF` already a float?

AngeloTadeucci: It was a lie. It was an coordS all the time

Zintixx (approved):

PR #637, opened 2022-05-25
Feat: Implement patrol data movement

AngeloTadeucci:
### Major
- Implemented NPC moving by **PatrolData**; **(It's not perfect yet, npcs will phase trough blocks)**
- Implemented trigger for walking character with **PatrolData**;
- Added **FollowNPC** packet;
- Merged **Mob** and **Npc** classes;
- Removed duplicated Mob/Npc packets. They had the same struct;
- Fixed/Updated Npc movement... again...

### Minor
- Reworked how **PatrolData** is parsed;
- Added **PatrolDataUuid** to **MapNpc** class;
- Simplified **FieldManager.RequestNpc** function;


MapleServer2/Managers/Actors/Npc.cs
```diff
@@ -1,30 +1,415 @@
...
+
     public override void Animate(string sequenceName)
     {
         Animation = AnimationStorage.GetSequenceIdBySequenceName(Value.Model, sequenceName);
         // TODO implement stopping animation
     }
+
+    public override void Damage(DamageHandler damage, GameSession session)
+    {
+        if (Value.Friendly == 2) // Attacking friendly NPC
```
AlanMorel, line 72: I wonder if `2` should be an enum so it's readable that it is "attacking"

MapleServer2/Managers/Actors/Npc.cs
```diff
@@ -1,30 +1,415 @@
...
+    /// </summary>
+    /// <returns>True if the NPC is moving.</returns>
+    public bool UpdateVelocity()
+    {
+        if (WayPointQueue.Count > 0 && Distance <= 0)
+        {
+            return GetNextPath();
+        }
+
+        if (Distance <= 0)
```
AlanMorel, line 105: Rather than check `Distance <= 0` twice, try checking it once, then immediately check the `Count`, if it's greater than `0`, return that, otherwise return false.

MapleServer2/Managers/Actors/Npc.cs
```diff
@@ -1,30 +1,415 @@
...
+        if (difference.Length() < 0.1f)
+        {
+            SetMovementTargetToDefault();
+            return;
+        }
+
+        int tickNow = Environment.TickCount;
+        int timeDelta = Math.Clamp(tickNow - LastMovementTime, 0, 400);
+
+        const float UnitPerMs = 0.001f; // Velocity in which all NPCs move per ms. 150 units per second.
```
AlanMorel, line 138: I feel like this could also be made a constant somewhere in the codebase

AlanMorel (changes requested):

AlanMorel (approved):

darkvergus (approved):

AlanMorel (approved):

Zintixx (approved):

PR #660, opened 2022-06-02
Fix: enchanting

AngeloTadeucci:
- Fix event id for ophelia/peachy. (It was my fault, oops)
- Fix: null exception

### DIFF
--- a/Maple2Storage/Scripts/Npcs/11000508.lua
+++ b/Maple2Storage/Scripts/Npcs/11000508.lua
@@ -8,25 +8,25 @@ function getBeginEventId(itemRarity, enchantLevel)
             return 1001
         end
     end
-    return 1998 + enchantLevel
+    return 1198 + enchantLevel
 end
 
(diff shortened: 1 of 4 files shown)


AlanMorel (approved):

Maple2Storage/Scripts/Npcs/11000508.lua
```diff
@@ -8,25 +8,25 @@ function getBeginEventId(itemRarity, enchantLevel)
             return 1001
         end
     end
-    return 1998 + enchantLevel
+    return 1198 + enchantLevel
```
Sparkymod, line 11: i dislike the idea of hardcode numbers, but aparently is the best option right now

Sparkymod (approved):

Maple2Storage/Scripts/Npcs/11000508.lua
AngeloTadeucci: its not hardcoded numbers, its the the event ids from the XMLs.

Zintixx (approved):

PR #671, opened 2022-06-16
Item Parser and Metadata clean up

Zintixx:
- Clean up and reorganized the item parser. Parsed a few more attributes
- Removed `IsTemplate` bool for Items. Now checks if item is UGC by mesh from metadata. By default, all UGC-able items create a placeholder UGC class.
- Implements the formula for instrument performance EXP and for prestige EXP.
- Implements multiple checks for using or equipping items.


AngeloTadeucci (approved):
Good stuff

PR #687, opened 2022-06-22
Fix: Mounts now consume stamina when sprinting

### LINKED ISSUE #650: Mount sprinting doesn't consume stamina
How to reproduce:
- Give yourself any mount
- Hold CTRL
- You can sprint infinitely

https://user-images.githubusercontent.com/106274631/170841009-d4c2bc43-b65a-41c0-a3f2-6161f5fdd946.mp4

The debug console will still send packets for Stamina Consumption but the server won't receive them:
image

gleefuldev:
Fixes #650 

The current amount that the stamina consumption is set to isn't exactly the same as in the original servers, but I believe this is unable to be fixed at the moment because of how stamina regeneration currently works.

### DIFF
--- a/MapleServer2/Constants/Constant.cs
+++ b/MapleServer2/Constants/Constant.cs
@@ -3,4 +3,5 @@
 public static class Constant
 {
     public const float UnitPerMs = 0.001f; // Velocity in which all NPCs move per ms. 150 units per second.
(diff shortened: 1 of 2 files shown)


AngeloTadeucci (approved):
Noice!

darkvergus (approved):

PR #689, opened 2022-06-23
Character Creation Checks, Login Time Fix, WIP Mushking Royale Additions

Zintixx:
- Added some additional checks upon character creation
- Fixed timestamps on characters when logging in
- Basic foundations for Mushking Royale sessions and queue
  - With this comes some additions to triggers


MapleServer2/Types/MushkingRoyaleQueue.cs
```diff
@@ -0,0 +1,68 @@
...
+                    StartMatch();
+                }
+                await Task.Delay(10000); // 10 second delay
+            }
+        });
+    }
+
+    private void StartMatch()
+    {
+        List<Player> players = Players.Take(50).ToList();
```
AngeloTadeucci, line 59: I think using an actual queue here is best , at least you will get the players in order of join which is nice

AngeloTadeucci (approved):

MapleServer2/Types/MushkingRoyaleQueue.cs
Zintixx: Doesn't Take do this?
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.take?view=net-6.0

newer players in the queue would be much lower in the queue

AngeloTadeucci: Yeah I guess it didn't matter

AlanMorel (approved):

PR #714, opened 2022-07-03
Feat: Wardrobe

### LINKED ISSUE #336: Wardrobe UI
Character Page related, but none of this is functioning apart from opening the UI itself.

MapleStory2_FCbdf2M4Tu

Zintixx:
- Implements wardrobe
- Resolves #336 
- Moved item equip/unequip to Inventory.
- Shuffled how ItemSlots work in order to take into consideration wardrobe.


MapleServer2/Types/Inventory.cs
```diff
@@ -359,6 +359,100 @@ public void MoveItem(GameSession session, long uid, short dstSlot)
...
+            return false;
+        }
+
+        // Remove the item from the users inventory
+        RemoveItem(session, uid, out item);
+
+        Player player = session.Player;
+
+        // Get correct equipped inventory
+        Dictionary<ItemSlot, Item> equippedInventory = player.GetEquippedInventory(item.InventoryTab);
```
AngeloTadeucci, line 381: I think I added some bind handling around here, what happened to it?

AngeloTadeucci (changes requested):

AngeloTadeucci (approved):

mettaursp (approved):

MapleServer2/PacketHandlers/Game/WardrobeHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private static void HandleRename(GameSession session, PacketReader packet)
+    {
+        int index = packet.ReadInt();
+        string name = packet.ReadUnicodeString();
+
+        if (!session.Player.TryGetWardrobe(index, out Wardrobe wardrobe))
+        {
+            _ = new Wardrobe(0, 0, index, name, new(), session.Player);
```
Clockworkx, line 140: I think there should be an explicit function call to assign the wardrobe to the player, instead of using the constructor for that.
I would prefer these two function calls instead of this line.
or a function that calls these two, either would be good I think.

       DatabaseManager.Wardrobes.Insert(this, player.CharacterId);
        player.Wardrobes.Insert(index, this);

I dont like calling the Wardrobe constructor for that.

MapleServer2/Types/IInventory.cs
```diff
@@ -72,6 +72,8 @@ public interface IInventory
     IEnumerable<Item> GetAllByFunctionId(int functionId);
     Item GetFromInventoryOrEquipped(long uid);
     bool ItemIsEquipped(long uid);
+    bool Equip(GameSession session, long uid, ItemSlot slot);
```
Clockworkx, line 75: TryEquip to signal that it can fail?
Same for the function below?

MapleServer2/PacketHandlers/Game/WardrobeHandler.cs
AngeloTadeucci: I think the database insert inside the constructor it makes sense but it should add to player while creating like player.Wardrobe.Insert(new Wardrobe())

Clockworkx (approved):

Sparkymod (approved):

PR #719, opened 2022-07-05
Fix: Server crashing after making a character or closing client

craftingmod:
Added some safe checking nulls of `SkillTab.cs` and `GameSession.cs`

# The reason of crashing
1. When making character, subSkill in SkillTab.cs might be null.
2. Sometimes, when you are closing client OnlineCTS might be null.

### DIFF
--- a/MapleServer2/Servers/Game/GameSession.cs
+++ b/MapleServer2/Servers/Game/GameSession.cs
@@ -61,7 +61,7 @@ public class GameSession : Session
         FieldManager.RemovePlayer(Player);
         GameServer.PlayerManager.RemovePlayer(Player);
 
-        Player.OnlineCTS.Cancel();
+        Player.OnlineCTS?.Cancel();
         Player.OnlineTimeThread = null;
 
         CoordF safeCoord = Player.SafeBlock;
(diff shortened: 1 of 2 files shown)


MapleServer2/Servers/Game/GameSession.cs
```diff
@@ -61,7 +61,7 @@ protected override void EndSession(bool logoutNotice)
         FieldManager.RemovePlayer(Player);
         GameServer.PlayerManager.RemovePlayer(Player);
 
-        Player.OnlineCTS.Cancel();
```
AngeloTadeucci, line 64: Just by creating a new character and not logging in can cause this line to reach? I don't think that's possible, since creating a character uses LoginSession 🤔

AngeloTadeucci (approved):

MapleServer2/Servers/Game/GameSession.cs
craftingmod: That OnlineCTS NullReferenceError is from disconnecting game (In login) after creating new character.

I forgot `when playing in character and` prefix. 😭

AngeloTadeucci: Sounds good!

Zintixx (approved):

Issue #721, opened 2022-07-05
Project style


AngeloTadeucci:
### Handlers
   - All must implement LogUnknownMode if they have multiple modes.
   - All modes must be separated into different methods.
   - All must be stateless.

### Packets
   - Only use one file per Send/Recv OP.
   - All mode must have their own method with an enum for the mode byte.

## GameDataParser
   - Don't import MapleServer2 into GameDataParser project.
   - Metadatas must be readonly or init-only. You can use record for that.
  
## Maple2Storage
   - Files that can be used by GameDataParser and MapleServer2 should live here. (e. g.: Metadata files)

PR #723, opened 2022-07-06
Consistent mode naming and handler/packet name changes

Zintixx:
- Removed request/response in "RequestXHandler"/"ResponseXPacket". It's pretty obvious its a request and response at this point. This should hopefully keep some consistency with all the other handlers and packets
- Renamed all modes to be "Mode" rather than a descriptor of the mode within the handler/packet. Again, this is also pretty obvious the mode is for that specific instance.
- Updated some packets to be static for consistency
- #721

### DIFF
--- a/MapleServer2/Enums/BuildModeType.cs
+++ b/MapleServer2/Enums/BuildModeType.cs
@@ -1,0 +1,8 @@
+﻿namespace MapleServer2.Enums;
+
+public enum BuildModeType : byte
+{
+    Stop = 0,
+    House = 1,
+    Liftables = 2
+}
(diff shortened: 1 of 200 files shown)


AngeloTadeucci:
Rename missing on:
- MapleServer2/PacketHandlers/Game/RequestCubeHandler.cs
- MapleServer2/PacketHandlers/Game/LiftableHandler.cs

AngeloTadeucci (approved):

darkvergus:
Ok what kinda of wizard are you? How is this compiling in the first place? You're not using the new enum you made for BuildModeType in LiftableHandler. Zin trying to break master once more :pepelaugh:

darkvergus (approved):

PR #729, opened 2022-07-13
More triggers

AngeloTadeucci:
- #508 
- Related to soul binder tutorial


AlanMorel (approved):

Clockworkx:
In general I am not sure whether the word Relevant in getRelevantQuests is descriptive, or useful for understanding the function better, if whatever we are getting would not be relevant, we would not use the function,
there is also no getIrrelevantQuests :D, so it does not seem a useful distinction

Zintixx (approved):

PR #782, opened 2022-10-17
Mob AI Infra v0.1

AngeloTadeucci:
This is the base of the new Mob AI. It works similar to triggers where each AIState has an OnEnter, Execute and OnExit

Example on how to implement new AI:
- MapleServer2/AIScripts/PrisonGovernor2Battle.cs

Behavior applied to all mobs:
- MapleServer2/AIScripts/DefaultNew.cs

To define which AI an npc will use, the index of AI is here:
- MapleServer2/AIScripts/AiHelper.cs


MapleServer2/AIScripts/PrisonGovernor2Battle.cs
```diff
@@ -0,0 +1,71 @@
...
+using MapleServer2.Data.Static;
+using MapleServer2.Managers.Actors;
+using MapleServer2.Types;
+
+namespace MapleServer2.AIScripts;
+
+public class PrisonGovernor2Battle : AIState
+{
+    public override void OnEnter(Npc npc)
+    {
```
Zintixx, line 11: In the same sense that a field has Triggers and TriggerLoader, I feel as though we should do the same for NPCs, so we wouldn't have to feed in the NPC instance to get NPC variables and values.

MapleServer2/AIScripts/PrisonGovernor2Battle.cs
```diff
@@ -0,0 +1,71 @@
...
+        // Ok to assume this will only happen on tutorial, so always attack the only player available
+        npc.Target = npc.FieldManager!.State.Players.FirstOrDefault().Value;
+    }
+
+    public override AIState? Execute(Npc npc)
+    {
+        if (npc.CheckHPThreshold(50))
+        {
+            return new BattleStop();
+        }
```
Zintixx, line 21: Likewise with this. `CheckHPThreshold` would be an AI function and not necessarily under NPC, like how we have `TriggerCondition` all in one area specifically for trigger scripts

MapleServer2/AIScripts/PrisonGovernor2Battle.cs
```diff
@@ -0,0 +1,71 @@
...
+        if (playerMobDist >= npc.Value.NpcMetadataDistance.Avoid * 2)
+        {
+            npc.Animation = AnimationStorage.GetSequenceIdBySequenceName(npc.Value.NpcMetadataModel.Model, "Run_A");
+
+            // Get coord between the player and the mob that is the avoid distance
+            CoordF coord = CoordF.From(
+                npc.Target.Coord.X + (npc.Coord.X - npc.Target.Coord.X) * npc.Value.NpcMetadataDistance.Avoid / playerMobDist,
+                npc.Target.Coord.Y + (npc.Coord.Y - npc.Target.Coord.Y) * npc.Value.NpcMetadataDistance.Avoid / playerMobDist,
+                npc.Target.Coord.Z + (npc.Coord.Z - npc.Target.Coord.Z) * npc.Value.NpcMetadataDistance.Avoid / playerMobDist
+            );
```
Zintixx, line 39: This is better but I think these could also be AI functions like `CheckHPThreshold`

MapleServer2/AIScripts/PrisonGovernor2Battle.cs
```diff
@@ -0,0 +1,71 @@
...
+    public override AIState? Execute(Npc npc)
+    {
+        PlayerTrigger? playerTrigger = npc.Target!.Value.Triggers.FirstOrDefault(y => y.Key == "battleStop");
+        if (playerTrigger is not null)
+        {
+            playerTrigger.Value = 1;
+            return null;
+        }
+
+        npc.Target.Value.Triggers.Add(new("battleStop", 1));
```
Zintixx, line 66: This should all be in an AI function, named something like `SetUserValue`

Essentially what I'm getting at with all these is that logic of each AI script shouldn't be in the AI script. This should all be in one/multiple AI files (like how we have Trigger functions).

Having logic per script would increase the tediousness of creating each script. It's like equivalent as if trigger scripts for maps had logic in them. They don't. Instead, they just have conditions, actions, and their respective parameters.

Zintixx (changes requested):

MapleServer2/AIScripts/AiHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        {
+            "AI_DefaultNew.xml",
+            new DefaultNew()
+        },
+        {
+            "AI_PrisonGovernor2_battle.xml",
+            new PrisonGovernor2Battle()
+        }
+    };
```
Zintixx, line 17: hmm is there a way to automate this? Could we just parse the ai script name from the NPC xmls without file extension and name the AI script exactly as it is?
Just thinking ahead when there's dozens and dozens more.

Maple2Storage/Types/Metadata/NpcMetadata.cs
```diff
@@ -19,7 +19,7 @@ public class NpcMetadata
     [XmlElement(Order = 6)]
     public byte Level;
     [XmlElement(Order = 7)]
-    public Dictionary<NpcState, (string, NpcAction, short)[]> StateActions = new();
+    public Dictionary<NpcState, (string id, NpcAction action, short chance)[]> StateActions = new();
```
Sparkymod, line 22: you can consider use a class or record at this stage rather than this generic, like NpcStateAction and have those 3 fields there.

Sparkymod (changes requested):

MapleServer2/AIScripts/DefaultNew.cs
```diff
@@ -1,10 +1,29 @@
...
 using MapleServer2.Data.Static;
 using MapleServer2.Enums;
 using MapleServer2.Managers.Actors;
 using MapleServer2.Types;
 
-namespace MapleServer2.AIScripts.Default;
+namespace MapleServer2.AIScripts;
+
+public class DefaultNew : AIState
+
```
AlanMorel, line 11: Extra blank line

AlanMorel (changes requested):

AlanMorel (approved):

MapleServer2/AI/Functions/AICondition.cs
```diff
@@ -0,0 +1,60 @@
...
+    public bool InSight()
+    {
+        if (Npc.Target is null)
+        {
+            return true;
+        }
+
+        float playerMobDist = GetDistanceToTarget();
+        return playerMobDist < Npc.Value.NpcMetadataDistance.Sight * 2;
+    }
```
Zintixx, line 34: Couldn't this just be 1 function?

Zintixx (changes requested):

MapleServer2/AI/Functions/AICondition.cs
AngeloTadeucci: Oops, fixed. auto pilot

Zintixx (approved):

Sparkymod (approved):

Issue #797, opened 2022-10-27
Warehouse gold coin storage error


Maplems:
[2022-10-27 22:56:42] [Fatal] MapleServer2.Network.Session: Exception in recv PipeScheduler: System.NullReferenceException: Object reference not set to an instance of an object.
   at MapleServer2.Types.Currency.UpdateWallet() in D:\Maple2\MapleServer2-master\MapleServer2\Types\Currency.cs:line 61
   at MapleServer2.Types.Currency.Modify(Int64 input) in D:\Maple2\MapleServer2-master\MapleServer2\Types\Currency.cs:line 27
   at MapleServer2.PacketHandlers.Game.ItemStorageHandler.HandleMesos(GameSession session, PacketReader packet) in D:\Maple2\MapleServer2-master\MapleServer2\PacketHandlers\Game\ItemStorageHandler.cs:line 113
   at MapleServer2.PacketHandlers.Game.ItemStorageHandler.Handle(GameSession session, PacketReader packet) in D:\Maple2\MapleServer2-master\MapleServer2\PacketHandlers\Game\ItemStorageHandler.cs:line 41
   at MapleServer2.Network.PacketRouter`1.OnPacket(Object sender, PacketReader reader) in D:\Maple2\MapleServer2-master\MapleServer2\Network\PacketRouter.cs:line 23
   at MapleServer2.Network.Session.ReadRecvPipe(PipeReader reader) in D:\Maple2\MapleServer2-master\MapleServer2\Network\Session.cs:line 263
[2022-10-27 22:56:42] [Information] MapleServer2.Servers.Game.GameServer: Game client disconnected: GameSession from 127.0.0.1:53681

PR #798, opened 2022-10-27
Fix: Add Session to BankInventory's Currency

Zintixx:
- Fixes: #797


AngeloTadeucci (approved):

AngeloTadeucci:
I really wish we could remove Session or Player from Currency, but I don't know how.

Issue #761, opened 2022-09-19
Some epic quests do not progress.


Kustale:
During some quests, such as The great winn stilton, there is a map in which the character changes, but the moment I proceed to this map, an error occurs on the server (however, the server does not shut down). And all functions except for shutting down whether the connection is lost. And when I reconnect, it enters the map that keeps having that problem and keeps getting errors. (There may be problems with some other quests, but I haven't found them yet)

(Sorry, I would like to write the code of the error, but I will update the error code later due to current computer repair
And, bad english)

* I have temporarily resolved the above issue. Access the SQL tool and change the map code to 2000001(Tria) in Database - Character - Map (?). And when I reconnected, I escaped (However, infinite fall T_T: In this case, find the Kerning City map code in the chat window and enter /m. Then find the quest code and enter /completequest)

AngeloTadeucci:
What are the errors that happens when you enter the map? We need that information!

Kustale:
(Forcibly restarting this quest.  /startquest 60100115 true
And, it's very long to write more details about this error.
Server version is September 13th)

1. Enter the door in Dark Wind HQ
2. This is where Lennon and Winn Stilton fall down.
3. The story begins (Lennon and Katvan's men fight), and Katvan appears at the end.
4. After the story. map changed. (Map name is Dark Wind HQ Passage)
5. Here, the character should be changed to Lennon, but it does not change, only my character. And I stop here and cannot proceed. (Disconnected from server)

An error occurred on the server and then hung up.
The content of that error:

[20:11:35] [ERR] FieldManager: System.Collections.Generic.KeyNotFoundException: The given key '10000' was not present in the dictionary.
   at System.Collections.Concurrent.ConcurrentDictionary`2.ThrowKeyNotFoundException(TKey key)
   at System.Collections.Concurrent.ConcurrentDictionary`2.get_Item(TKey key)
   at MapleServer2.Triggers.TriggerContext.SetSound(Int32 soundId, Boolean isEnabled) in D:\MapleServer2-master\MapleServer2\Triggers\FieldObjectTriggers.cs:line 182
   at Maple2.Trigger._52000138_qd._main.StateAct1_wave1.OnEnter()
   at MapleServer2.Triggers.TriggerScript.Next() in D:\MapleServer2-master\MapleServer2\Triggers\TriggerScript.cs:line 25
[...]

Issue #756, opened 2022-09-16
Null reference with gemstone


Maplems:
[2022-09-16 18:36:12] [Fatal] MapleServer2.Network.Session: Exception in recv PipeScheduler: System.NullReferenceException: Object reference not set to an instance of an object.
   at MapleServer2.Types.GemSockets.CreateSockets(Item parent) in D:\Maple2\MapleServer2_20220916\MapleServer2\Types\Gemstone.cs:line 89
   at MapleServer2.Types.GemSockets..ctor(Item parent) in D:\Maple2\MapleServer2_20220916\MapleServer2\Types\Gemstone.cs:line 41
   at MapleServer2.Types.Item..ctor(Int32 id, Int32 amount, Int32 rarity, Boolean saveToDatabase) in D:\Maple2\MapleServer2_20220916\MapleServer2\Types\Item.cs:line 144
   at MapleServer2.Managers.Actors.Npc.HandleMobKill(GameSession session, Npc npc) in D:\Maple2\MapleServer2_20220916\MapleServer2\Managers\Actors\Npc.cs:line 339
   at MapleServer2.Managers.Actors.Npc.Damage(DamageHandler damage, GameSession session) in D:\Maple2\MapleServer2_20220916\MapleServer2\Managers\Actors\Npc.cs:line 108
   at MapleServer2.PacketHandlers.Game.SkillHandler.HandleDamage(GameSession session, PacketReader packet) in D:\Maple2\MapleServer2_20220916\MapleServer2\PacketHandlers\Game\SkillHandler.cs:line 266
   at MapleServer2.PacketHandlers.Game.SkillHandler.HandleDamageMode(GameSession session, PacketReader packet) in D:\Maple2\MapleServer2_20220916\MapleServer2\PacketHandlers\Game\SkillHandler.cs:line 160
   at MapleServer2.PacketHandlers.Game.SkillHandler.Handle(GameSession session, PacketReader packet) in D:\Maple2\MapleServer2_20220916\MapleServer2\PacketHandlers\Game\SkillHandler.cs:line 34
   at MapleServer2.Network.PacketRouter`1.OnPacket(Object sender, PacketReader reader) in D:\Maple2\MapleServer2_20220916\MapleServer2\Network\PacketRouter.cs:line 23
   at MapleServer2.Network.Session.ReadRecvPipe(PipeReader reader) in D:\Maple2\MapleServer2_20220916\MapleServer2\Network\Session.cs:line 263

PR #799, opened 2022-10-28
Fix: nullables

AngeloTadeucci:
- Fix: #756 
- Fix: #761

### DIFF
--- a/MapleServer2/Types/Gemstone.cs
+++ b/MapleServer2/Types/Gemstone.cs
@@ -13,8 +13,8 @@ public class Gemstone
     public string OwnerName = "";
     public bool IsLocked;
     public long UnlockTime;
-    public ItemStats Stats;
-    public ItemAdditionalEffectMetadata AdditionalEffects;
+    public ItemStats? Stats;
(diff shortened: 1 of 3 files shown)


AlanMorel (approved):

MapleServer2/Types/Gemstone.cs
```diff
@@ -108,7 +111,7 @@ public void CreateSockets(Item parent)
     public void OpenSockets()
     {
         // roll to unlock sockets
-        for (int i = 0; i < Sockets.Count; i++)
+        foreach (GemSocket socket in Sockets)
         {
             int successNumber = Random.Shared.Next(0, 100);
```
Zintixx, line 117: Was this tested? I remember specifically using a for loop instead of foreach loop for a reason.

Zintixx (changes requested):

MapleServer2/Types/Gemstone.cs
AngeloTadeucci: No but how would it make a difference?

Zintixx (approved):

